PR #6, opened 2015-08-02
BLUETOOTH -- Rebased for Scott

Wackerbarth:
This is the effort that @thinkyhead made to make the references to Bluetooth less cryptic


thinkyhead:
Instance names should be lower-case. Otherwise, looks good.

Wackerbarth:
Rebased and merged

PR #7, opened 2015-08-02
Provide both long and short version strings

Wackerbarth:
PR moved from Marlin


boelle:
travis does not check this one... but is it safe to merge anyway?

i guess that closing and submitting again will make travis check it

AnHardt:
@ boelle
No, it's a conflict with #24. Both of us changed the same area of code.

boelle:
oki, just thinking loud here :-P

PR #8, opened 2015-08-02
[Concept Only] A way to use Arduino Menus in Configuration Selection

Wackerbarth:
This code is "concept only" and, intentionally incomplete.


boelle:
could the incomplete also be the reason for the conflicts?

boelle:
just sav that travis did check this one....

PR #9, opened 2015-08-02
Servo pin fix for Azteeg X3 Pro

Natealus:
Recent changes made old conditionals fault in the X3 Pro pin file. So I
removed the #if statements and just directly defined the servo pins and
servo system works as usual now.


AnHardt:
The problem occures not only for Azteeg X3 Pro.
It's verified at least for all RAMPS-boards to. I suppose we need this for __all__ pin-files.

thinkyhead:
The `#undef` statements are redundant and can be removed. Servo pins are undefined until the board-specific pins file is processed.

To work around the problem where `HAS_SERVOS` isn't defined, a better solution is to move the definition of `HAS_SERVOS` above the line in `Conditionals.h` just before `#include "pins.h"`.

thinkyhead:
#11 takes this to the extreme. The code doesn't need to count servos when making the pin definitions because the conditionals throughout the code handle that.

PR #33, opened 2015-08-03
Fix AutoBedLevel Bug

setar:
Fix merged with MarlinDev


boelle:
hmm the build failed... checking why

boelle:
if you click the details link above (https://travis-ci.org/MarlinFirmware/MarlinDev/builds/73887891) and go to line 305 it might give you a clue....

@AnHardt & @Wackerbarth you have any idea why it fails?

AnHardt:
A missing ')' right before the last ';' in line 2747 of Marlin_main.cpp.

But anyway, i'd not merge it until i understand what it's doing.

setar:
sorry,
I did a test compilation in a different directory (MarlinFirmware/Marlin)

Wackerbarth:
" I'd not merge it until i understand what it's doing." -- I totally agree. At the moment, bed leveling is our biggest problem. It seems close to working and we need to be VERY CAREFUL about any changes.

That is not to say that there is anything "wrong" with this. In fact it is likely to fix some of our problems. BUT, we need a much better understanding of both the WHAT and WHY of this submission.

Wackerbarth:
@setar -- (Marlin repository) That's interesting. Both repositories are the same if you are referencing dev (and Development isn't far behind) Since the error was an unbalanced ")", I'm surprised that it works in one and not the other.  Perhaps your configuration failed to actually include the code. Did your compile Configuration specify a DELTA configuration?

setar:
https://github.com/MarlinFirmware/Marlin/issues/2040#issuecomment-116496471
main idea:

To determine the coordinates of the point located in the rotated coordinate system in the real coordinate system used rotation matrix apply_rotation_xyz (plan_bed_level_matrix, x, y, z);

To get the real form the bed in the rotated coordinate system, it is necessary to carry out an inverse transform using an inverse matrix.

This code measures the height of the bed in the rotated coordinates, finds the minimum and maximum, and makes a correction based on the fact that the current position of the head corresponds to the last probe point

setar:
@Wackerbarth 
at the moment I have not tested configuration DELTA, because they do not have such hardware.

AnHardt:
As the other changes in that area it makes no sense to test it with a delta.
The correction is only applied for "#if DISABLED(DELTA)"

Wackerbarth:
The correction is only applied for "#if DISABLED(DELTA)"  -- Yeh, I realized that after I ask.
But, I will argue that if this correction is really needed, then it should apply to deltas also.


I guess that just helps point out how unreadable our code is. We really need to break it down into manageable pieces.

AnHardt:
At least i got now that, in the special case of a rotation matrix, the transposed is the same as the inverse matrix.
All in all it looks promising. If give it a try on hardware tonight.

If it works as expected - we'll have a look on the deltas.

Wackerbarth:
I see the explanation, but who has actually tested this?
If we want to bring code into our repository, we shouldn't be adding it to "master" until it has had some testing.

Perhaps we need to reconsider our workflow.

boelle:
i assume that @setar did a test since it his PR

boelle:
but also if we have to consider our workflow we just end up as before.... nothing happens for weeks...

creating a new repro and then stalling on merge PR's then we are not better of then before....

Wackerbarth:
Yes, that is assumed. But I think we need an additional verification. No reasonable developer is going to submit something that doesn't work as he expects.  But, since there is no design specification, how do we know that there isn't something that he might not have considered.

Wackerbarth:
@boelle -- You are missing the point. "master" needs to be stuff that we know we want to keep for the next release. We cannot just dump everything in without having some consideration of its impact.

Right now, we are trying to get out a 1.1 Release Candidate.  I have no idea if this code helps, or hurts, that effort.

A workable workflow merges in things that are working. It doesn't try to extract "the good stuff" from some mishmash of this and that.

Personally, I consider that we have been using the individual fork repositories as the holding place for pending code. Posting it as a PR gives notice that it needs to be considered.

An alternate workflow would be to create many branches within MarlinDev and open a new branch for each PR.

boelle:
well i'm just tired of all the talking and we are still waiting for the ok on auto level

i tried it but it did not raise the Z between probes resulting in my probe being dragged sideways over the bed.

i had to use a fork pointed out by @Roxy-3DPrintBoard to get it going. and its still not perfect but i can use it as is

Wackerbarth:
And how does your using Roxy's fork help us get 1.1 working? If you are going to start throwing in a bunch of new stuff, we can't possibly figure out what is wrong with the current code. You're asking us to shoot at a moving target. And this latest piece of code just moved the CRITICAL target again.

thinkyhead:
> it did not raise the Z between probes

We should troubleshoot when we find a machine that actually exhibits the problem. Most of them seem to be working now, making it harder to find testers for the bugs.

AnHardt:
https://github.com/MarlinFirmware/MarlinDev/pull/40#issuecomment-127611892

PR #40, opened 2015-08-03
Revert "Fix AutoBedLevel Bug"

AnHardt:
@setar 
It would be nice if youd test your code at least a bit.

These loops can not terminate because yy can never be less than zero (negative) because of uint8_t.
```
for (uint8_t yy = auto_bed_leveling_grid_points - 1; yy >= 0; yy--) {
```


Wackerbarth:
@boelle-- This is EXACTLY why you need to slow down a little. NOTHING should go into Master until someone, in addition to the author, has actually reviewed/tried it.

thinkyhead:
On a side-note, I would like to try and preserve the comments describing the final position correction, because this is an area that has been confusing in the past and while we are still scrubbing `G29` issues I would like it to be there as a signpost for contributing developers to be cautious and to know why it is the way it is.

setar:
@AnHardt Thanks 
"This is the result of optimization of the advice ..."

Colleagues, I do not have extensive experience in programming, especially the collaboration.
But the logic of the code works, and fixes a big problem that has to date.

AnHardt:
To be honest, i can't see what problem is solved here.

What problems do we have when the probing process is working at all?
a) "I see different print heights between G28 and G29."
    Is the really a problem? No. If you have a fixed workflow you ether do a G28, or a G28 G29.
    Who cares about the unused workflow.
   Let's assume it is a problem.
   Them the observation is incomplete. A more complete observation is. "I see different print heights at
    the same coordinate, depending on if i used G28 and G29." Well, that's the nature of the thing.
    If we use G28 only, we see different distances between nozzle an bed because of the slope and the
    warp of the bed, depending on the coordinate we look at it. That's the reason for we do a correction.
    If we use G29, we see different distances between nozzle an bed because of the the warp of the
    bed, depending on the coordinate we look at it. That looks pretty much the same, but the
   average failure is smaller. (That doe's not necessary mean that the maximum an minimum
    differences are smaller, but normally they are) So yes, the observation is correct, but lies in the
    nature of the thing and is the wanted effect.

b) "My nozzle is scratching the bed." or air-printing.
    With a simple G28 the answer to repair this is to `adjust the endstop`.
   And with a G29? It's the same! Here we call it `zprobe_zoffset` and we can adjust it on the fly.
    We do not need any further corrections!

c) "I don't want to make a G29 all the time."
    Let's store the rotation matrix in the EEPROM and home z at the same point as the last probe of
   G29.

    We do not need a additional z-correction.

setar:
You're right, this is the basic code to implement the following algorithm: the gradual compensation of a curved surface to the plane.
I have invented the algorithm now I code it.
if it works, I'll post the next patch

PR #45, opened 2015-08-04
add support FSR sensors

setar:
draft code for add support FSR sensors as Z probe


Wurstnase:
Would it not be better to use new files for that?
I've started it on my branch like this:
https://github.com/Wurstnase/Marlin4Due/blob/master/Marlin/fsr_sensor.h

thinkyhead:
@setar I suggest doing a rebase against the latest MarlinDev/master, then taking a diff and applying to a clean copy of MarlinDev/master as a new branch, then finally cherry-picking to make sure only your new feature is included. Then start a new PR, which will be a bit cleaner and not include any accidental reversions that don't apply. Unfortunately the latest changes to master are making this into a little mess.

setar:
OK, I'll do it later

boelle:
+1 from me... been out to cool off from my rant about things going slow

but this one... i really like it as it makes the mechanical side much more simple... why use a servo etc when a single FSR can do it

oh i hear people scream about plastic on the nozzle.... really simple... just clean it before you do an G29... and you dont have to do a G29 before each print....

i use a steel feeler blade like the ones you use to adjust spark plugs with... put a 0.2mm one under the nozzle and make it grap it... heat the nozzle and jog up 0.1-0.2 and slide the blade back and fort and sideways... clean the blade and repeat... 

0.2 is also the figure that most nozzle expands when heating up

but again i will use this as soon i get a new extruder block made

thinkyhead:
I'm re-working this to create a new PR, making some corrections to the language and so forth. One thing I notice is the presence of the new setting `FSR_OFFSET_FROM_EXTRUDER` and I think this should be eliminated. For one thing, probing the bed should not set the Z position. Only when __homing__ with a probe does Marlin care how far it is from the nozzle, and we already have a setting for that, `Z_PROBE_OFFSET_FROM_EXTRUDER`. I also wonder if you can just add FSR support into `run_z_probe` rather than creating a separate function for it.

Wackerbarth:
Incorporated in PR#81

Issue #43, opened 2015-08-04
SD: Init caught in endless loop when init fails


Sniffle:
as the description says SD:init goes into an endless loop when an SD card is not inserted.  

I inserted an SD card to test if the loop continued and it does not.  

This is on the rep rap discount full graphics controller

Wackerbarth:
I've run into the same thing.  I narrowed it down to Aug 2 to the interval between 8d3b74c and eb65c57 where all of the SDCARDDETECT stuff was changed.

With no card in my BrainwavePro board, everything seemed fine until then.
But after the change, I get a constant stream of alerts and the system seems to have slowed down because of the extra overhead.

Sniffle:
I also experiences the slow down from the extra overhead

Wackerbarth:
Fixed!

thinkyhead:
Sorry, I neglected to test this with the SD card removed.

PR #46, opened 2015-08-04
Restore test of SD Card present

Wackerbarth:
This fixes issue #43

PR#2572 Changed this because of faulty Boolean Logic

NOT(A OR (NOT B) OR (NOT C)) !=

(NOT A) AND ((NOT B) OR C))

Besides, the compiler should be smart enough to
optimize this without help from the programmer


thinkyhead:
Darnit, I thought really hard and that logic seemed so good in my head. I must think harder next time. And, I was trying to improve clarity for humans, not to optimize ahead of the compiler.

Issue #48, opened 2015-08-05
G29 E broken since switch to MarlinDev, likely servo related?


Natealus:
I've gone over a plethora of variables to try and see what the root cause is. Even tried the new bed leveling code that hasn't been merged yet and same result. It has something to do with the servo system I believe because it just moves to the first spot and engages the servo up and down about 5-10 times, might move a slight bit then goes to the next line as it climbs upwards to a crash with the Y tower. Offsets, number of points, servo tweaks don't seem to be the affecting variables. I've pull fresh copies and rewritten the configuration.h files 2 times to make sure I got all of the changes over. 

Note: I pulled the delta/generic example configs so if there is something not changed there to match the new enable/disable system that could be an issue.

Natealus:
It would be really helpful if we could use the suggestion I had in issue #12 about a configuration option to enable debugging output. Especially in regards to ABL.

Natealus:
Maybe we should bring some back story into it. Back when I installed the hardware a couple of months ago when delta bed leveling was being implemented...G29 E is the only way I could get the probing array to probe all the way through successfully. If I use M401 to pre-extend the z-probe and just use G29 V4 then it goes down and does 0 to 2 probes successfully then stops on the next probe press and makes a weird sound...its not a stepper skip but it sounds something rough for a second then it stops and gives me a sequence like this.

->G29 V4

<-G29 Auto Bed Leveling
<-Bed X: 46.000 Y: -74.000 Z: 11.775
<-Bed X: 22.000 Y: -74.000 Z: 16.762
<-Bed X: -2.000 Y: -74.000 Z: 19.500
<-Bed X: -26.000 Y: -74.000 Z: 24.487
<-Bed X: -50.000 Y: -74.000 Z: 29.475
<-Bed X: -74.000 Y: -50.000 Z: 34.462
<-Bed X: -50.000 Y: -50.000 Z: 39.450
<-Bed X: -26.000 Y: -50.000 Z: 44.437
<-Bed X: -2.000 Y: -50.000 Z: 49.425
<-Bed X: 22.000 Y: -50.000 Z: 54.412
<-Bed X: 46.000 Y: -50.000 Z: 59.400
<-Bed X: 70.000 Y: -50.000 Z: 64.388
<-Bed X: 94.000 Y: -26.000 Z: 69.375
<-Bed X: 70.000 Y: -26.000 Z: 74.363
<-Bed X: 46.000 Y: -26.000 Z: 79.350
[...]

PR #54, opened 2015-08-05
Debug G29 -- As a full fledged debug option

Wackerbarth:
To use this feature, you must build the firmware with ```DEBUG_LEVELING_FEATURE``` enabled.
Then, to activate it during a run, use
```
G28
M111 S32
G29
```


AnHardt:
For sure this does not improve the readability of the code.

Wackerbarth:
I agree that this detailed debugging info makes the code less readable.
However, that is what is required to make it both a compile time and also, thereafter, a runtime option.

I feel that we need to have both of those aspects. Compile time to avoid wasting ROM and Run time to have a system that doesn't spit out reams of garbage that does not apply to the user's current activity.

Natealus:
Hmm as I go to compile the debug_g29 branch with or without the #define DEBUG_LEVELING_FEATURE, it throws me some compiling errors. What machine setup do you have @Wackerbarth ? I use a delta setup.

Arduino: 1.6.5 (Windows 8.1), Board: "Arduino Mega or Mega 2560, ATmega2560 (Mega 2560)"

Marlin_main.cpp: In function 'void gcode_SET_Z_PROBE_OFFSET()':
Marlin_main.cpp:5221: error: 'zprobe_zoffset' was not declared in this scope
         zprobe_zoffset = value;
         ^
In file included from Marlin_main.cpp:30:0:
Marlin_main.cpp:5239: error: 'zprobe_zoffset' was not declared in this scope
       SERIAL_ECHO(zprobe_zoffset);
                   ^
Marlin.h:67:43: note: in definition of macro 'SERIAL_PROTOCOL'
 #define SERIAL_PROTOCOL(x) MYSERIAL.print(x)
                                           ^
Marlin_main.cpp:5239:7: note: in expansion of macro 'SERIAL_ECHO'
       SERIAL_ECHO(zprobe_zoffset);
       ^
'zprobe_zoffset' was not declared in this scope

Wackerbarth:
@Natealus -- I'm also running a Delta.
I suspect that there is some difference in your Configuration.h and mine.  I didn't update all of the various Configuration.h files to reflect this change, just the default one (which has Auto bed leveling off by default)

Make sure that you have the following
```
//===========================================================================
//============================ Bed Auto Leveling ============================
//===========================================================================

// @section bedlevel

#define AUTO_BED_LEVELING_FEATURE // Delete the comment to enable (remove // at the start of the line)
#define DEBUG_LEVELING_FEATURE
//#define Z_PROBE_REPEATABILITY_TEST  // If not commented out, Z-Probe Repeatability test will be included if Auto Bed Leveling is Enabled.
```

Natealus:
Oh vwallah I looked more closely then noticed you made a commit that changed (ENABLE_AUTO_BED_LEVELING) to (AUTO_BED_LEVELING_FEATURE)

Natealus:
Here's one where it went and probed a few points correctly...would messup on a point then continues in a diagonal Z fashion. Eventually screws z down till the servo gets too low and starts skipping. Manually turned it off.

http://pastebin.com/ycAFSxzG

Natealus:
@Wackerbarth @AnHardt I moved the debugs over to issue #48 
If there are more conditions you think I should try, I'm happy to test it out I'm on debug mode right now :)

Wackerbarth:
Yes, I saw them come in.  Unfortunately, I am much too sleepy to make any sense of them. (0200 local, and I am an early riser rather than a night owl). I'll look at them in the morning. If we cannot see the motion that you describe, I found another place where we can extract some positioning information. So I may need to add some code and have you re-run these same sequences.

Question: Are the behaviors reproducible? (If you try the same test sequence, do you get the same behavior?)

Natealus:
@Wackerbarth Indeed a good question. No they are not the same each time, sometimes it chooses to probe in the first spot many times then moving on sometimes the first couple probes work correctly then it starts screwing up. Makes...confusing debugging. 

But no I totally understand being tired and code becoming a blur. Sleep is a great thing I hope you get a good rest tonight. :)

thinkyhead:
I never really intended this to be deployed. I just needed to get a very detailed view of these procedures so I could diagnose issues remotely. (Since I don't have a leveling probe myself.) At the moment I still need to rebase / merge my own debug_G29 branch with the latest MarlinDev/master, but the logging code included in this PR should be all the same as mine.

thinkyhead:
> sometimes it chooses to probe in the first spot many times then moving on sometimes the first couple probes work correctly then it starts screwing up

@Natealus Classic symptoms of buffer overrun, data corruption, signed/unsigned integer mismatch, etc. Bugs like this are tricky to find. The debug output should give some clue.

Natealus:
@thinkyhead Always with the tricky problems I bring. lol but if there's some specific conditions you want me to check let me know and I'll give it a debugging whirl. I did try to reduce my buffers to see if free memory was the issue but I have 2,667 bytes of ram free for local variables now and still the same.

thinkyhead:
@Natealus I have a new debug branch based on code from yesterday (8-9-2015) at https://github.com/thinkyhead/MarlinDev/tree/debug_G29

Try `G29` with that and post the output. Point out where in the log output it's being weird, and we should be able to use that to find the specific spot in the code where it's messing up.

Natealus:
@thinkyhead I just did a clean reconfigure of the latest MarlinDev master branch but it throws me errors saying I don't have a Z_MIN_PROBE_PIN defined when...it does. I'm not sure where it's getting redirected cause I even tried #undef and #define Z_MIN_PROBE_PIN and it still tells me

Arduino: 1.6.5 (Windows 8.1), Board: "Arduino Mega or Mega 2560, ATmega2560 (Mega 2560)"

In file included from Configuration_adv.h:608:0,
                 from Configuration.h:973,
                 from Abacus3D-MarlinDev.ino:33:
SanityCheck.h:150: error: #error You must have a Z_MIN_PROBE_PIN defined in your pins_XXXX.h file if you enable Z_MIN_PROBE_ENDSTOP.
         #error You must have a Z_MIN_PROBE_PIN defined in your pins_XXXX.h file if you enable Z_MIN_PROBE_ENDSTOP.
          ^
#error You must have a Z_MIN_PROBE_PIN defined in your pins_XXXX.h file if you enable Z_MIN_PROBE_ENDSTOP.

Is it a SanityCheck.h error? cause it's there in pins_AZTEEG_X3_PRO.h from AnHardt's commit.

Wackerbarth:
@Natealus -- "econfigure of the latest MarlinDev master branch" 
"from Abacus3D-MarlinDev.ino:33:"
There is no Abacus3D-MarlinDev.ino in the master branch and "latest" is somewhat ill defined.
Would you please turn ALL of your changes into a commit based on whatever version of "master" you are using and push it to a public repository so that we can fetch THE EXACT FILES that you are attempting to compile?

Natealus:
@Wackerbarth Well yeah I make a separate folder for my personal printer firmware. When pull requests are merged I copy the affected files or changes in the config into Abacus3D-MarlinDev

If I make changes in MarlinDev folder (IE configuration for my printer) GitHub for Windows wants to block the program up saying I can't merge those commits (which it shouldn't cause that configuration is specifically for my printer not base configuration.h) and it won't download any new pulls that were merged.

I'm just trying to do a basic setup, not make any new commits at this time. But I guess making a fork of what I currently am trying to compile would help you identify the issue?

Wackerbarth:
@Natealus -"making a fork of what I currently am trying to compile would help you identify the issue?

Exactly! It's MUCH easier to read the lines in your files rather than guessing what you might have put in there.

Natealus:
@Wackerbarth Shall do so. Newer version of the Github program might work better for this sort of thing. Previously it was troublesome to use the same Marlin folder for the printer config cause it wanted to merge them to the MarlinFirmware/Marlin. Now it compares forks better.

Natealus:
@Wackerbarth Here we are good sir,
https://github.com/Natealus/Abacus3D-MarlinDev/
The only changes are configuration.h and configuration_adv.h from the generic delta examples + my settings.

Wackerbarth:
Well, I compared to my configuration and the critical difference is that I DO NOT DISABLE_MIN_ENDSTOPS.  Turning them back on will allow it to compile.

However, it SHOULD accept the Z_MIN_PROBE,  Unfortunately, for deltas, that does not work at the present time.

Natealus:
Ah that makes some sense cause they're trying to merge it to Z_MIN in some ways cause apparently if you use probe that technically is a Z_MIN anyways.

Natealus:
Hmm that doesn't seem to be the fix for my setup. I wonder if it's a false sanitycheck fail?

thinkyhead:
@Natealus I'll have to look at the min probe error issue later tonight or tomorrow, as I'm on a bus to Canada at the moment. If you post your configurations then I can use them for my testing.

Natealus:
@thinkyhead Cool are you taking a trip for fun or visiting family? Yeah as I check it again and again it seems to be pinned in right...just wondering if it was a false positive sanitycheck.

PR #55, opened 2015-08-05
Reverse x-axis direction of H-bot feature

AnHardt:
Add the define to enable reverse x-axis direction for core bots to Configuration.h
Make the motion equation dependend on that in planner.cpp

Hopefully as suggested in :https://github.com/MarlinFirmware/Marlin/issues/2587
@DWONH could you please test this.

If all is well i'll distribute the changes to the other configs.


Marlin/planner.cpp
```diff
@@ -540,13 +540,21 @@ float junction_deviation = 0.1;
     // corexy planning
     // these equations follow the form of the dA and dB equations on http://www.corexy.com/theory.html
     block->steps[A_AXIS] = labs(dx + dy);
-    block->steps[B_AXIS] = labs(dx - dy);
     block->steps[Z_AXIS] = labs(dz);
+    #if DISABLED(CORE_REVERSE_X_AXIS_FEATURE)
```
Wackerbarth: I suggest that you change this to #if ENABLED(...) and reverse the following blocks.

At least to me, it would be easier to follow.

Marlin/planner.cpp
```diff
@@ -540,13 +540,21 @@ float junction_deviation = 0.1;
...
+    #else
+      block->steps[B_AXIS] = labs(-dx - dy);
+    #endif
   #elif ENABLED(COREXZ)
     // corexz planning
     block->steps[A_AXIS] = labs(dx + dz);
     block->steps[Y_AXIS] = labs(dy);
-    block->steps[C_AXIS] = labs(dx - dz);
+    #if DISABLED(CORE_REVERSE_X_AXIS_FEATURE)
+      block->steps[B_AXIS] = labs(dx - dy);
```
Wurstnase: Where is the C_AXIS gone? and ```dz```?

Wurstnase:
I would suggest another solution.
When you switch the Pins in the pins.h for X and Y or maybe also Z you can change any axis. 

Maybe
```
#define X_DIR_ORG 19 // the original pin

#if ENABLED(SWITCH_SOMETHING)
  #define Y_DIR X_DIR_ORG
  #define Y_DIR_INV !X_DIR_INV_ORG
#else
  #define X_DIR X_DIR_ORG
.
.
.
```

Wackerbarth:
I certainly like this (Wurstnase’s) approach better.
It can all be migrated to the HAL and doesn’t affect the fundamental kinematic routine.

DWONH:
I tested @whosawhatsis suggested, but it didn't work. same result as before.
The problem here is that changing directions in H-bot(corexy) needs changing two motors together, and it reverse other axis' direction too.

DWONH:
also you need to modify below line in stepper.c

    if ((((out_bits & (1<<X_AXIS)) == 0)&&(out_bits & (1<<Y_AXIS)) == 0)) {   //-X occurs for -A and -B

to

    if ((((out_bits & (1<<X_AXIS)) != 0)&&(out_bits & (1<<Y_AXIS)) != 0)) {   //-X occurs for +A and +B

Unless it can not check endstop hit.

I could change X-direction by changing code(equation part & above). 
but I saw strange bug during test. checkHitEndstops() function is called during printing (echo: endstops hit: Y?). I'm testing now but I can not reproduce it anymore. :(

Wackerbarth:
In CORE XY

A = X + Y
B = X - Y

In H

a = -X + Y
b = -X - Y

Therefore

b = -A
a = -B

So you need to switch and invert both of these 

#define X_DIR_ORG 19 // the original pin
#define Y_DIR_ORG ?? // the other original pin

#if ENABLED(SWITCH_SOMETHING)
  #define Y_DIR X_DIR_ORG
  #define Y_DIR_INV !X_DIR_INV_ORG
  #define X_DIR Y_DIR_ORG
  #define X_DIR_INV !Y_DIR_INV_ORG
#else
  #define X_DIR X_DIR_ORG
  #define X_DIR_INV X_DIR_INV_ORG
  #define Y_DIR Y_DIR_ORG
  #define Y_DIR_INV Y_DIR_INV_ORG
#endif

grob6000:
Why not just configure an H-bot by swapping the axes (either in firmware or just relabel the motors and swap the cables!!) and inverting both directions (or just flip the motor cables!!)? Do we really need another config option here?

DWONH:
@AnHardt I'm not familiar with GitHub system, what's meaning of closing pull request? Will new option for this provided in next release?

Wurstnase:
Do it like @Wackerbarth said. Just switch your Motors on your RAMPS or whatelse and invert the direction and you are done.

thinkyhead:
Responding to this and to https://github.com/MarlinFirmware/Marlin/issues/2587#issuecomment-128593100 …

I realize that there is a hardware solution. And yet, we include `INVERT_[XYZ]_DIR` configuration options, also easily done by flipping the plug. So I think it's right and consistent to include these same options (or interpret them differently) when using `COREXY` and `COREXZ`.

In terms of mechanics, the `INVERT_[XYZ]_DIR` options assume that "normally" the motors will turn clockwise as the X or Y coordinate increases, but when you want them to turn counter-clockwise as X or Y increases you change the setting.

With `COREXY` we now have a different setup, where two motors combine to produce movement. The "normal" in that case is for the motors to move clockwise when Axis A/B coordinates increase. When `COREXY` is enabled the `INVERT_[XYZ]_DIR` settings actually apply to Axis A, B, Z, and with `COREXZ` the affected axes are A, Y, C.

So, I think Richard is correct, that the solution should be based on the `INVERT_[XYZ]_DIR` settings, but we need to interpret them differently with `COREXY` and `COREXZ`.

Wackerbarth:
In terms of mechanics, the INVERT_[XYZ]_DIR options assume that "normally" the motors will turn clockwise as the X or Y coordinate increases, but when you want them to turn counter-clockwise as X or Y increases you change the setting.

Yes, you can either change the physical wire or the "virtual" one to invert a direction.

The same thing is true about swapping motors. You can change the cable, swapping the physical location of the motor controlled by a particular wire, or you can change which signal you send down that wire by swapping the designated pin numbers.

Similarly, you can swap the "increases both X and Y" motor for the "increases X but decreases Y" motor.

So, in reality, COREXY and COREXZ are really no more that a similar swapping of the Y and Z axes.

thinkyhead:
I always prefer to keep the plugs un-flipped so it's easy to remember how to re-attach them if unplugged. So I always opt to use the software setting. I see that since the invert options always apply to individual motors (via the implementation) _and_ since this may be desirable even for core machines, that any option to reverse core direction is best done as it is in this PR. So… I wouldn't mind seeing this PR fixed up (to allow reversal of either of the two-motor axes) and merged.

whosawhatsis:
It should be possible to implement this in software without adding any code. You can switch the motors by changing pin assignments, and while I haven't looked closely at the how CoreXY is currently implemented, it SHOULD still be possible to use the existing direction inversion options to reverse the directions of the motors so that you don't need to flip the plugs. If this is true, it's possible to implement this change within the current configuration without hardware changes, so the proposed change is a little redundant.

Wackerbarth:
@whosawhatsis -- I COMPLETELY agree. Changing the pin numbers has the same effect as moving a wire from one plug to another. And inverting the stepper direction is the same as swapping the coil leads on a particular stepper.

It is unfortunate that the legacy naming of the steppers uses the same designation as that of the Cartesian Object Coordinates (XYZ) rather than that of some other system (ABC) to which they really belong.

In the Cartesian case, WE SHOULD label the motors as DIRECT-A, DIRECT-B, and DIRECT-C.

In the case of COREXY, COREXZ and this COREH configuration, the motors need to be labeled SUM-A, DIFFERENCE-B, and DIRECT-C.

In the delta case, we have DELTA-A, DELTA-B, and DELTA-C.

This PR needs to be closed as "UNNEEDED" and the issue becomes one of "DOCUMENTATION".

AnHardt:
This PR ___is___  closed since a long time. Labeled the original issue as  'Documentation Issue'.

mojtaba-hdpr:
i have problem with this !!
anyone solve this problem that x axis in h-bot is revers ??

Wackerbarth:
@mojtaba-hdpr -- You can fix your problem by physically swapping and/or inverting the stepper motor wires. Alternatively, you can accomplish the same thing by changing the pin assignment/direction settings in your configuration file.  Read the description supplied earlier in this thread.

mojtaba-hdpr:
in equation of corexy in marlin A=x+y and B=x-y but in one of model h-bot ( for example in this pic " https://cloud.githubusercontent.com/assets/13655997/9081538/fe9085e0-3b97-11e5-968b-090be116c675.png ") A=x+y and B=y-x
 if i change x direction or x-step motor A=y-x and B=y+x
and in real axis X and Y to Y and X ,
in repetier firmware is 2 format for corexy one B=y-x and one B=x-y;
in total we have 4*2=8 options for corexy that :
1)A=x+y ;   2)A=-(x+y) ;    3)A=x-y ;       4)A=y-x ;
   B=y-x ;       B=x-y ;            B=-y-x ;         B=y+x ;
these options changing between's with revers x or y or x and y axis with wiring;
but if we need to change for one of any options that i write in bottom we need to change marlin cod !!
5)A=x-y ;   6)A=-(x-y) ;    7)A=x+y ;       8)A=y+x ;
   B=y-x ;       B=x-y ;            B=-y-x ;         B=y+x ;
please attention to equations !!!!

Wackerbarth:
Left | Right | Step1 | Step2
------------ | ------------- | ------------- | -------------
 X+Y  |    X-Y   |  A  |  B
 X+Y  |   -X+Y  |  A  | -B
 X-Y   |   X+Y   |  B  |  A
 X-Y   |  -X-Y  |  B  | -A
-X+Y  |    X+Y | -B |   A
-X+Y  |   -X-Y  |  -B  | -A
-X-Y   |   X-Y  | -A  |  B
-X-Y   |  -X+Y | -A  | -B

To change A to -A you change 
#define INVERT_X_DIR true

To swap the A & B columns, you change the pin numbers

There is no need to change the code,
it is all in the configuration.

thinkyhead:
General H-Bot motion equations for reference. From http://corexy.com/theory.html
```
∆X = (∆A + ∆B) ÷ 2
∆Y = (∆A - ∆B) ÷ 2
∆A = ∆X + ∆Y
∆B = ∆X - ∆Y
```

PR #57, opened 2015-08-06
QUICKHOME for COREXY/XZ

darkjavi:
Coming from https://github.com/MarlinFirmware/Marlin/pull/2378
Reworked to include helper functions to compare floats, not sure if macros.h is the right place for those functions though.
I did it also for the CoreXZ as I think it could be also exposing the issue.
 
Test results:
https://youtu.be/CYL-U1YJ5jY

Once again I would like to ask CoreXY/XZ owners to test it  @Salandora  @ruggb.

Also I can confirm what @ruggb pointed out in the old thread, I noticed there is no stepper timeout, as far as 20 minutes the drivers kept active.


Marlin/stepper.cpp
```diff
@@ -309,13 +309,23 @@ inline void update_endstops() {
...
   #define TEST_ENDSTOP(ENDSTOP) (TEST(current_endstop_bits, ENDSTOP) && TEST(old_endstop_bits, ENDSTOP))
 
-  #define UPDATE_ENDSTOP(AXIS,MINMAX) \
-    SET_ENDSTOP_BIT(AXIS, MINMAX); \
-    if (TEST_ENDSTOP(_ENDSTOP(AXIS, MINMAX))  && (current_block->steps[_AXIS(AXIS)] > 0)) { \
-      endstops_trigsteps[_AXIS(AXIS)] = count_position[_AXIS(AXIS)]; \
-      _ENDSTOP_HIT(AXIS); \
-      step_events_completed = current_block->step_event_count; \
-    }
+  #if (DISABLED(COREXY) && DISABLED(COREXZ)) // non hbot / CoreXY / CoreXZ
```
Wackerbarth: I suggest that this be inverted to #if (ENABLED(COREXY) || ENABLED(COREXZ))
 ...... The core version
#else 
---- The others
#endif

Salandora:
Well Sorry I'm out, I had to switch my Electronics and went over to a Due based one.

ruggb:
@darkjavi 

Yes, I currently have about a 2-3wk old dev running but forgot about chkg the driver issue. Previously, once it was enabled with any motion command it stayed activated forever. I think the setting I have now turns it loose immediately - which is not good. I guess i will have to recheck things.
Is what u want tested in the current dev dnld?

darkjavi:
I'm doing my best to keep up with the feedback I'm getting for the PRs, but looks like I still have a lot to learn about the insights of the avr programing.
Let's talk about the issue before doing another pr; The thing is that the ENDSTOP_UPDATE macro is checking if the axis the endstop is mounted on is moving or it won't update the endstop status.

```C
if (TEST_ENDSTOP(_ENDSTOP(AXIS, MINMAX))  && (current_block->steps[_AXIS(AXIS)] > 0))
```
This is been working correctly for DELTA and regular cartesian printers, in which each stepper can only hit the endstop of his axis. 

CoreXY is a little bit different though, as it can move diagonally in XY using only one stepper, and I think both 'regular home' and quickhome has been bugly working in two ways:

1-Regular home- Moving just X(or Y) causes both steppers(A/B) to move at the same time, so when it comes to check for steps[_AXIS(AXIS)] > 0 it will be true as both A&B have steps, but the intent is to check for X/Y not A/B.

2-QuickHome- Quick home exposes a problem for the machines in which X_MAX == Y_MAX, as the initial diagonal move will be performed with only one stepper(A), thus it will crash the head if not hitting the X endstop first.

For the machines in which X_max != Y_max it will bugly work for the same reason as the 'regular homing' as the first diagonal move wont be a 45º diagonal thus it will use both steppers(AB) to perfom it.



Thinking about it I can come with 3 different approaches:

1- Disabling quickhome for corexy, so it will keep working bugly for the reasons explained above.

2- Removing the check for steps in the UPDATE_ENDSTOPS, so it will look like:
```C
if (TEST_ENDSTOP(_ENDSTOP(AXIS, MINMAX)) 
```
This was my first approach in https://github.com/MarlinFirmware/Marlin/pull/2255. Don't know if it has any futher implications but it worked.

3- Reworking this PR approach so it will use boolean instead of floats to tell if the axis is moving, something like:

In planner.h replace
```C
float axis_movement[NUM_AXIS];
```
```C
bool is_axis_moving[NUM_AXIS];
```
load the variable in planner.cpp and then in the UPDATE_ENDSTOP macro:
```C
if (TEST_ENDSTOP(_ENDSTOP(AXIS, MINMAX))  && (current_block->is_axis_moving[_AXIS(AXIS)]))
```
This way it will only check for a bool in interrupt time instead of a float.


Sorry for the brick, feedback and approaches are welcome :)

Marlin/stepper.cpp
darkjavi: Here I was trying to follow the scheme I saw in different places in the code which is : 

if !exception
 General case
else
  Particular case

Wackerbarth: Yes, I understand that you were following a pattern used elsewhere.

However, that reminds me of Lemmings and the sea.

The reason *I* do not like that form is that when reading it, it is far too easy to fail to see, or at least recognize, the negation.  Most people think   IF ... THEN in positive terms. Further the negated form goes against the CASE construct
```
IF Condition1 THEN Action1
ELSE IF Condition2 THEN Action2
OTHERWISE DefaultAction
```
But it is really a style issue. In this case, the actual code performs the intended action in each case.

AnHardt: Many of this cases come from replacing #ifndef with DISABLED(). #ifndef looks pretty normal to me, while DISABLED() looks as if the logic can be improved. :-)

Readers - don't care.
Writers - change i you work close to it and feel disturbed.

thinkyhead:
@darkjavi Good catch on the `ENDSTOP_UPDATE` macro and how it interacts badly with `COREXn` machines.

thinkyhead:
I'll chime on on this minor point…
```
if !exception
  General case
else
  Particular case
```
The reason I prefer the opposite is that it puts the General Case at the end, closer to the code that follows, making it easier to see the General Case _as a whole_ on a single screen.

Wackerbarth:
@thinkyhead -- Just to clarify,
we both prefer
```
if exception
  Particular case
else
  General case
```

Correct?

thinkyhead:
@Wackerbarth Yes. The general case last.

thinkyhead:
@darkjavi I'd still like to merge this if it can be tweaked as suggested. We've needed the `nearly_zero` macro for some time now, not to mention `block->axis_movement` values for non-cartesian robots. It will be cleanest to re-roll this in a new branch with new commits, then submit a new PR rather than adding new commits to correct previous ones, but do whatever works for you.

Issue #37, opened 2015-08-03
Error Recovery Problem


Wackerbarth:
Because of positional errors, at the end of bed leveling, the attempt to retract the probe failed.
```
> 12:38:26 PM: N831 M111 S6 *77
< 12:38:26 PM: FIRMWARE_NAME:Marlin 1.1 rkw-163-g1869e90 SOURCE_CODE_URL:https://github.com/Wackerbarth/MarlinDev/ PROTOCOL_VERSION:1.0 MACHINE_TYPE:Kossel Pro EXTRUDER_COUNT:1 UUID:00000000-0000-0000-0000-000000000000
> 12:38:26 PM: N832 M111 S4 *76
> 12:38:38 PM: N844 G28 X0.00 Y0.00 Z0.00 *94
> 12:38:46 PM: N853 G29 *60
< 12:39:09 PM: -4.09 -4.23 -4.18 -3.94 -4.26 -4.19 -3.67 
< 12:39:09 PM: -3.97 -4.01 -4.05 -4.07 -4.03 -3.74 -3.42 
< 12:39:09 PM: -3.86 -3.89 -3.92 -3.87 -3.80 -3.59 -3.39 
< 12:39:09 PM: -3.66 -3.76 -3.81 -3.77 -3.76 -3.68 -3.38 
< 12:39:09 PM: -3.47 -3.49 -3.51 -3.54 -3.44 -3.31 -3.19 
< 12:39:09 PM: -3.27 -3.28 -3.28 -3.34 -3.21 -3.06 -2.91 
< 12:39:09 PM: -3.08 -3.08 -3.06 -2.93 -2.98 -2.88 -2.68 
< 12:39:14 PM: Error:Z-Probe failed to retract!
< 12:39:14 PM: Error:Printer stopped due to errors. Fix the error and use M999 to restart. (Temperature is reset. Set it after restarting)
> 12:40:07 PM: N906 M999 *5
< 12:40:07 PM: Resend: 855
< 12:40:08 PM: Error:Line Number is not Last Line Number+1, Last Line: 854
< 12:40:08 PM: Resend: 855
< 12:40:09 PM: Error:Line Number is not Last Line Number+1, Last Line: 854
< 12:40:09 PM: Resend: 855
```
However, this issue is that I issued the M999 command to resume.
But, it failed because the line numbers are now out-of-sync between the printer and the host driving it.

thinkyhead:
> N906 M999 *5

[...]

PR #60, opened 2015-08-07
Continue from the line following the M999

Wackerbarth:
If the printer is going to accept the M999,
then it should continue from that line number.

Fixes issue #37

### DIFF
--- a/Marlin/Marlin_main.cpp
+++ b/Marlin/Marlin_main.cpp
@@ -5083,7 +5083,7 @@ inline void gcode_M907() {
 inline void gcode_M999() {
   Running = true;
   lcd_reset_alert_level();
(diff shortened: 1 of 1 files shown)


thinkyhead:
Cool. So is `M999` only allowed from hosts or does it apply in some way to SD too? I guess SD wouldn't even have line numbers so this fix should be ok if it does.

thinkyhead:
I see that this is the only place at all that `Stopped_gcode_LastN` is ever used, so that variable can be dropped altogether if it's not useful for future features.

Wackerbarth:
True.  I left everything around only because someone might have a use case and I didn't want to purge it prematurely.

However, as I understand the current compilers, it will probably get optimized away even if we leave the code there.

PR #62, opened 2015-08-08
Rename Z_RAISE_BEFORE_PROBING and Z_RAISE_AFTER_PROBING to Z_PROBE_SE…

AnHardt:
Rename `Z_RAISE_BEFORE_PROBING` and `Z_RAISE_AFTER_PROBING` to `Z_PROBE_SERVO_CLEARANCE` because they serve the same purpose.

Rename `Z_RAISE_BEFORE_HOMING` and `Z_RAISE_BETWEEN_PROBINGS` to `Z_RAISE_FOR_OBSTRACLES` because they serve the same purpose.

Kill a small bug in `raise_z_for_servo()` where `Z_RAISE_BEFORE_HOMING` was used instead of `Z_RAISE_BEFORE_PROBING`.


Marlin/Configuration.h
```diff
@@ -494,14 +494,11 @@ const bool Z_PROBE_ENDSTOP_INVERTING = false; // set to true to invert the logic
...
-  #define Z_RAISE_BEFORE_HOMING 4       // (in mm) Raise Z before homing (G28) for Probe Clearance.
-                                        // Be sure you have this distance over your Z_MAX_POS in case
-
   #define XY_TRAVEL_SPEED 8000         // X and Y axis travel speed between probes, in mm/min
 
-  #define Z_RAISE_BEFORE_PROBING 15   //How much the extruder will be raised before traveling to the first probing point.
-  #define Z_RAISE_BETWEEN_PROBINGS 5  //How much the extruder will be raised when traveling from between next probing points
-  #define Z_RAISE_AFTER_PROBING 15    //How much the extruder will be raised after the last probing point.
+  #define Z_PROBE_SERVO_CLEARANCE 15   // How much the z-axis will be raised before deploy/stove the servo probe.
+  #define Z_RAISE_FOR_OBSTRACLES  5    // Used in G28/G29 to be above the beds clamps/scrwes.
```
thinkyhead: Misspelling. Should be `OBSTACLES`.

Marlin/Configuration.h
```diff
@@ -494,14 +494,11 @@ const bool Z_PROBE_ENDSTOP_INVERTING = false; // set to true to invert the logic
...
+                                       // BE SURE TO HAVE THAT MUTCH SPACE TO THE TOP, WHEN STARTING A G28!
```
thinkyhead: Slightly better: `// Be sure to leave this much space at the top when starting a G28`

Marlin/Configuration.h
```diff
@@ -494,14 +494,11 @@ const bool Z_PROBE_ENDSTOP_INVERTING = false; // set to true to invert the logic
...
 
-  #define Z_RAISE_BEFORE_HOMING 4       // (in mm) Raise Z before homing (G28) for Probe Clearance.
-                                        // Be sure you have this distance over your Z_MAX_POS in case
-
   #define XY_TRAVEL_SPEED 8000         // X and Y axis travel speed between probes, in mm/min
 
-  #define Z_RAISE_BEFORE_PROBING 15   //How much the extruder will be raised before traveling to the first probing point.
-  #define Z_RAISE_BETWEEN_PROBINGS 5  //How much the extruder will be raised when traveling from between next probing points
-  #define Z_RAISE_AFTER_PROBING 15    //How much the extruder will be raised after the last probing point.
+  #define Z_PROBE_SERVO_CLEARANCE 15   // How much the z-axis will be raised before deploy/stove the servo probe.
```
thinkyhead: Misspelling: It should be `stow` (as in, put away) and not `stove` (as in, hearth, oven, or range).

thinkyhead:
I definitely like this PR – simplifies the issue and makes these settings useful for other things in future.

thinkyhead:
What else do we need to do to make `G30` sensible? I presume we can just add something like this…

    #if HAS_SERVO_ENDSTOPS
      raise_z_for_servo();
    #endif

Wackerbarth:
In the interest of "re-use", unification and simplification, rather than referring to SERVO this and SERVO that, shouldn't we just call it a PROBE  for example: MINIMUM_CLEARANCE_FOR _PROBE_DEPLOYMENT

thinkyhead:
@Wackerbarth Well, servos have been an exceptional case, because they have to rotate to deploy. Some are set up nicely, rotating to move an assembly straight up/down. Others just rotate an arm with a fat endstop switch at the end, and need the extra clearance specifically because of this feature.

thinkyhead:
How about instead of `Z_PROBE_SERVO_CLEARANCE` we call it `Z_PROBE_EXTRA_CLEARANCE`?

Wackerbarth:
I acknowledge that some "servos" need extra clearance. My point is that the more general concept is that the probe (whatever its design) needs certain clearances. For some designs, it may be as little as zero. But the concept is not limited to rotational servos.

For our documentation, we need a nice drawing of the probe with the various dimensions corresponding to the configuration parameters identified by the parameter name.

Wackerbarth:
Yes, Z_PROBE_EXTRA_CLEARANCE is preferable to Z_PROBE_SERVO_CLEARANCE.
But, just from the name, I don't know the point of reference from which that clearance is measured.

Sniffle:
It should be zprobe offset + z probe extra distance.

That way the z probe extra distance is always in addition to the extended
probe so that should the probe extra distance be called without being high
enough it will raise enough for the probe to be extended and still clear
any objects in the way.

Wackerbarth:
@Sniffle -- But we have to be careful. The last time I looked, even though the comment indicated otherwise, the probe offset was being specified as a negative number ( the coordinates of the probe in the nozzle frame of reference )

Personally, I prefer a positive value (the nozzle position in the probe reference frame) which is the height of the nozzle when the probe contacts the bed. In any case, I still advocate a diagram. A picture (in this case a simplified engineering drawing) really is worth many words.

Sniffle:
I completely agree to a drawing, im just saying that if the extra distance
should be in addition to the offset so that we raise the beight of the
hotend that much more.  Maybe a more proper expression would be hotend +
((abs value)offset+extra disance)(sign of offset?) so that we maintain to
move the hotend further away from the bed no matter its value negative or
positive

Wackerbarth:
AFAICT, we agree that we want the nozzle to be an amount that is in addition to the height obtained when the probe triggers.

And I think that your mathematical expression well demonstrates the need to keep both of those components positive in order to produce an understandable description.

AnHardt:
A servo probe:
servoprobe
Drawn when the tip of the switch (green) just meets the bed (red).
a) To actually trigger the switch we have to go down. This symbolized by the difference of the darkest to the medium gray circle. (For a inductive probe this is negative)
b) When we want to stow the servo clockwise, we have to list it by the difference of the medium and the light gray circle; else the switch case (blue) will hit the bed.
c) When we want to move the probe in x,y-direction we want to be higher then all the red stuff. We have to take into account screws and the slope of the bed.

So if we want to move in x,y-direction the minimal high is a+c. Z_RAISE_FOR_OBSTACLES
If we want to deploy or stow the probe the minimal height is a+b.  Z_PROBE_SERVO_CLEARANCE

Z_PROBE_OFFSET_FROM_EXTRUDER is the difference between nozzle-tip and the trigger point.

AnHardt:
Possibly we want to deploy the probe on top of the highest obstacle. Than we need a+b+c. Lets call it Z_PROBE_LEVEL or similar and use only that.

AnHardt:
How about:
```
#define Z_PROBE_LEVEL 10 // Minimal distance from the activated probe to the bed to activate, deactivate and operate the probe save. 
                         // Don't home if you don't have at least Z_PROBE_LEVEL mm to the top! Keep the value as small as possible.
```

AnHardt:
The wanted sequence for all probes than is:
* Raise before activating the probe.
* loop
*   Move to the probe XY.
*   Deploy the probe (if not deployed).
*   Do a Z probe at the current XY.
*   Raise Z before moving to the next XY.
*   (Optionally) Stow the probe.
* until all points are probed
* Stow the probe (if not stowed).

Ideally with a unified deploy/stow function for all types of probes (endstop/indictive, servo, sled, allen key, g-code sequence).
If you rally want to stow a allen key probe between the probing - do it - it's an option.

But i think this is still to complicated. We are doing to much. In reality we do not want the servo arm out of the way while moving, we want to re-position the servo at the beginning of a new probe. The servo may lose it's power after SERVO_DEACTIVATION_DELAY and the mechanical touch on the bed may change the servo-angle-position. Then simply refreshing the servo-angle-position is sufficient. (The repeatebility test is an other case?) Deploy/stow than can go out of the loop and we'd need a refresh() just before the actual z-probing. That way the delta-clipping-problem also would be solved.

Wackerbarth:
Your picture is a good start.  For clarity, let me suggest:
1) Use only one color for the probe body
2) Show a nozzle
3) Rather than shaded disks, use circles (colorless fill with visible border)
4) Crop off the top half (or more) of the circles
5) Use two or three illustrations to aid in explaining the WHY
6) Use dimensioning lines with the distances labeled (using their configuration name)
7) Make all configuration distances POSITIVE. Note that in some "unusual" cases (for example when the "probe" is actually the nozzle pressing on the bed) some values could be negative.
8) Rather than having the user compute the total height, I would add two components -- THE_HEIGHT_OF_THE_NOZZLE_WHEN_THE_PROBE_TRIGGERS and SOME_ADDITIONAL_PROBE_DEPLOYMENT_CLEARANCE. (Not those actual names, but shorter ones, as long as they retain clarity as to their meaning)

Also note that, when referring to "PROBE", "Z" and "MINIMUM" should be implied. So they can be omitted from any related names.

I'm guessing that you used scad to make the drawing.  If not, perhaps we should.
We should develop a library of components to use, and reuse, in making illustrations.

By storing them in the source tree under Documentation, we can start constructing illustrations with crude representations and allow contributors to help refine them.

Wackerbarth:
@AnHardt -- Yes, your state sequence is on the right track. I look forward to the re-implementation of these functions using concepts of this nature.

thinkyhead:
#75 is an update to this PR. The proposed `G30` changes are in a separate PR #73.

AnHardt:
Had some fun writing an Open-scad animation (A friend said "impossible!" - So i tried it.).
https://www.youtube.com/watch?v=w9RO8eRG2fU
Source at:
https://github.com/AnHardt/SCAD-Projects/blob/master/Servoprobe_bodycrash.scad
A bit crude, but works.
Just fill in the strings you want. ;-)

Wackerbarth:
@AnHardt -- Excellent. I would like for us to continue to refine the details and make it a permanent part of our documentation explaining bed leveling. My main suggestion is that we provide additional transition time between the various segments. From my brief look at your work, it should not be difficult to insert a few delays in your timeline.

thinkyhead:
@AnHardt I like it. I have corrected the text for you here, if you'd like to make a new video…
```scad
color([1.0,1.0,1.0,a1]) translate([-40,0,-44.3]) rotate([90,0,180]) text("_Null",5);
color([1.0,1.0,1.0,a4-a23]) translate([65,0,-51.7]) rotate([90,0,180]) text("Deploy_",5);
color([1.0,1.0,1.0,a7]) translate([-40,0,-47.7]) rotate([90,0,180]) text("_____Z_PROBE_OFFSET_FROM_EXTRUDER",5);
color([1.0,1.0,1.0,a9-a12]) translate([80,0,-70]) rotate([90,0,180]) text("There may be a screw",6);
color([1.0,1.0,1.0,a12-a17]) translate([80,0,-70]) rotate([90,0,180]) text("or a tilted bed",6);
color([1.0,1.0,1.0,a18-a23]) translate([-60,0,-55.2]) rotate([90,0,180]) text("_Travel_Height",5);
color([1.0,1.0,1.0,a22-a23]) translate([40,0,-56.9]) rotate([90,0,180]) text("Stow_",5);
color([1.0,1.0,1.0,a23]) translate([95,0,-56.9]) rotate([90,0,180]) text("Unified_Safety_Height_",5);
color([1.0,1.0,1.0,a24]) translate([80,0,-70]) rotate([90,0,180]) text("Regardless of the type of probe",6);
color([1.0,1.0,1.0,a24]) translate([80,0,-80]) rotate([90,0,180]) text("You can find a minimal safe height!",6);
```

Issue #50, opened 2015-08-05
Probe deploy/stow conditions for deltas only work when first and last points are inside DELTA_PROBABLE_RADIUS


AnHardt:
```cpp
          #if ENABLED(DELTA)
            // Avoid probing the corners (outside the round or hexagon print surface) on a delta printer.
            float distance_from_center = sqrt(xProbe*xProbe + yProbe*yProbe);
            if (distance_from_center > DELTA_PROBABLE_RADIUS) continue;
          #endif //DELTA

          ProbeAction act;
          if (deploy_probe_for_each_reading) // G29 E - Stow between probes
            act = ProbeDeployAndStow;
          else if (yCount == 0 && xCount == xStart)
            act = ProbeDeploy;
          else if (yCount == auto_bed_leveling_grid_points - 1 && xCount == xStop - xInc)
            act = ProbeStow;
          else
            act = ProbeStay;

          measured_z = probe_pt(xProbe, yProbe, z_before, act, verbose_level);
```
Maybe a decision based on probePointCounter is better. 
ProbeDeploy if probePointCounter == 0
ProbeStow if probePointCounter == auto_bed_leveling_grid_points*auto_bed_leveling_grid_points - skipped_points (+-1). 
With:
```
 if (distance_from_center > DELTA_PROBABLE_RADIUS) { skipped_points++; continue;  }
```

AnHardt:
No.
ProbeDeploy works, but ProbeStow is skipped (continue) again.

thinkyhead:
The solution is not to deploy on the first point, but on the first _allowed_ point. And the same with the last point. Meaning, it should stow at the end if it didn't already stow. The only "neat" way to do this is to add a new argument to `probe_pt` that indicates whether or not to _really_ probe. If the value is `true` then proceed normally, otherwise skip the `run_probe` part, but do the deploy/stow actions. If the action is `ProbeDeployAndStow` then don't _really_ deploy-and-stow unless also _really_ probing.

```cpp
bool really_probe = true;

#if ENABLED(DELTA)
  // Avoid probing the corners (outside the round or hexagon print surface) on a delta printer.
  float distance_from_center = sqrt(xProbe*xProbe + yProbe*yProbe);
  really_probe = (distance_from_center <= DELTA_PROBABLE_RADIUS);
#endif //DELTA

ProbeAction act;
if (deploy_probe_for_each_reading) // G29 E - Stow between probes
  act = ProbeDeployAndStow;
[...]

PR #63, opened 2015-08-08
After G29 - always end at the same point

AnHardt:
Always end at [RIGHT_PROBE_BED_POSITION, BACK_PROBE_BED_POSITION]
Regardless the evenness of auto_bed_leveling_grid_points.

Always zigzag
Use the shortest way between probing-points.

Add array indexIntoAB[xx][yy]
and use it to simplify the output loop when zigzaging.

### DIFF
--- a/Marlin/Marlin_main.cpp
+++ b/Marlin/Marlin_main.cpp
@@ -2565,7 +2565,7 @@ inline void gcode_G28() {
       #endif // !DELTA
 
       int probePointCounter = 0;
-      bool zig = true;
+      bool zig = (auto_bed_leveling_grid_points & 1) ? true : false; //always end at [RIGHT_PROBE_BED_POSITION, BACK_PROBE_BED_POSITION]
 
       for (int yCount = 0; yCount < auto_bed_leveling_grid_points; yCount++) {
         double yProbe = front_probe_bed_position + yGridSpacing * yCount;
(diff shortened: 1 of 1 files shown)


AnHardt:
As the code currently is as it is i have to start somewhere. Ending every time at the same point makes sense when in step two homing z will be done at this point. As long as your tracking and matrix code is not there this is absolute necessarily if we want to store the matrix for further use.

The delta clipping problem is an other one.  (#50) But i assume, when that one is solved it will be easy to add the save-homing-point (middle) to the probing sequence as the last point.

Marlin/Marlin_main.cpp
```diff
@@ -2621,12 +2616,13 @@ inline void gcode_G28() {
             eqnAMatrix[probePointCounter + 0 * abl2] = xProbe;
             eqnAMatrix[probePointCounter + 1 * abl2] = yProbe;
             eqnAMatrix[probePointCounter + 2 * abl2] = 1;
+            indexIntoAB[xCount][yCount] = probePointCounter;
```
thinkyhead: Or, is it possible to do this sort of thing instead…?
```cpp
typedef struct {
  float xindex;
  float yindex;
}  ProbePoint;
...
ProbePoint probePoint[auto_bed_leveling_grid_points * auto_bed_leveling_grid_points];
...
probePoint[probePointCounter].xindex = xCount;
probePoint[probePointCounter].yindex = yCount;
...
for (int ind = 0; ind < probePointCounter; ind++) {
  ???
}

thinkyhead:
Another naive question… If points are dropped because (for example) they don't fit within the circle of a delta bed, will the movement end up being strange?

PR #64, opened 2015-08-10
Filament sensor (all ramps configuration)

RicardoGA:
With this change the Filament sensor is available for all ramps configurations not only for EEB

### DIFF
--- a/Marlin/pins_RAMPS_13.h
+++ b/Marlin/pins_RAMPS_13.h
@@ -66,7 +66,7 @@
 #define SDSS               53
 #define LED_PIN            13
 
-#if MB(RAMPS_13_EEB) && ENABLED(FILAMENT_SENSOR)  // FMM added for Filament Extruder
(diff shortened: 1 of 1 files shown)


AnHardt:
The old condition
```
#if MB(RAMPS_13_EEB) && ENABLED(FILAMENT_SENSOR) 
```
makes me wonder. EEB is the dual extruder setup. It's used to activate a single filament width sensor.
EFB (the single extruder setup) would make more sense to me.
For all other numbers of extruders you'd need more then one filament width sensor.

RicardoGA:
that's right , the filament sensor is only for one extruder, but you can define the extruder number that has the filament sensor.

So you can use in any ramps configuration but only in one extruder in EEB or EEF.

clefranc:
@RicardoGA @AnHardt I've done some coding last winter for multi-sensor filament diameter, it's really simple to implement (build on top of what's existing right now), so perhaps on the next dev you'll implement this feature.

Issue #61, opened 2015-08-07
ABL delta for noobs


clefranc:
#### **A**uto **B**ed **L**eveling

---
**EDIT**: ***I've stopped at step 5 because of a logic flaw in delta homing when ABL is activated. A fix should be applied soon.***

The problem was exhibited by my *non-usual* probe; it's `TRIGGERED` when retracted. This has the side effect to mess homing (and even probing), because an endstop is immediately detected upon execution of the G28 command (home).

This is the drawback for using the probe as an endstop.

---
This is another attempt to use Marlin to operate a leveling probe on my delta printer. I've tried many times to make it work without success. The printer was originally using the Repetier firmware, and leveling was operational. So it must be doable with Marlin.

With a little help it should be running in no time. And I hope it'll help others make it work. Below are arbitrary steps that I've created, please feel free to comment or make suggestion.

---
**Step 1**: Download and unzip the latest code from https://github.com/MarlinFirmware/MarlinDev/tree/dev.

**Step 2**: Copy your delta configuration from ...\MarlinDev-dev\Marlin\example_configurations\delta  (in my case from biv2.5 folder) to D:\MarlinDev-dev\Marlin.

**Step 3**: Check that the Z_PROBE_PIN pin for the probe is defined in your pin board file, in my case pins_RUMBA.h:

```
#define Z_STEP_PIN         57
#define Z_DIR_PIN          56
#define Z_ENABLE_PIN       62
#define Z_MIN_PIN          -1 // Undefined to use as Z_PROBE_PIN.
#define Z_MAX_PIN          32

#define Z_PROBE_PIN        33 // Using the endstop Z_MIN_PIN.
```

**Step 4**: Adjust all endstops height (from Repetier ABL tutorial).

**Step 5**: ?

At this point, scrolling from Configuration.h, I stump first to the `Manual Bed Leveling` section.

What is manual bed levelling vs auto?
What does a mesh bed levelling do?

@thinkyhead  @Wackerbarth  What do you propose for this step? I think I need to test the probe with M119?

I'm going to bed (the real one after a night shift), I'll look at your comments when I wake up...

I'll update the OP with your help.

Thank you!

Wackerbarth:
Thanks. This is a good start to defining the kind of detail that we need to provide.
Obviously, the actual URL components will change as we move to different branches and repositories.

FYI, at the present time you should get CODE from the "dev" branch. (replace RCStaging with dev)
[...]

PR #65, opened 2015-08-10
Make Z_PROBE a Z_MIN_PROBE

AnHardt:
Currently we use the probe exclusively as a device to find the build platform (bed).
For the currently supported setups this means, we use it as an additional min-endstop.
A triggered-when-not-deployed probe disturbs the homing process for max-endstops.

- Rename `ENDSTOPPULLUP_ZPROBE` to `ENDSTOPPULLUP_ZMIN_PROBE`
- Rename `Z_PROBE_ENDSTOP_INVERTING` to `Z_MIN_PROBE_ENDSTOP_INVERTING`
- Rename `Z_PROBE_ENDSTOP` to `Z_MIN_PROBE_ENDSTOP`
- Rename `DISABLE_Z_PROBE_ENDSTOP` to `DISABLE_Z_MIN_PROBE_ENDSTOP`
- Rename `Z_PROBE_REPEATABILITY_TEST` to `Z_MIN_PROBE_REPEATABILITY_TEST`
- Rename `Z_PROBE_ENDSTOP` to `Z_MIN_PROBE_ENDSTOP`
- Adjust comments accordingly
- Remove Z_MAX  check for the probe in update_endstops().

Using a delta-related idea from @clefranc in #61,
extended to the general by @Wackerbarth to make the change for all setups.

Tested with Prusa i3, max-z-endstop and permanently-triggered z-probe.
Worked for @clefranc's delta.

### DIFF
--- a/Marlin/SanityCheck.h
+++ b/Marlin/SanityCheck.h
@@ -133,37 +133,37 @@
      * Require a Z Min pin
      */
     #if Z_MIN_PIN == -1
-      #if Z_PROBE_PIN == -1 || (DISABLED(Z_PROBE_ENDSTOP) || ENABLED(DISABLE_Z_PROBE_ENDSTOP)) // It's possible for someone to set a pin for the Z Probe, but not enable it.
-        #if ENABLED(Z_PROBE_REPEATABILITY_TEST)
-          #error You must have a Z_MIN or Z_PROBE endstop to enable Z_PROBE_REPEATABILITY_TEST.
+      #if Z_MIN_PROBE_PIN == -1 || (DISABLED(Z_MIN_PROBE_ENDSTOP) || ENABLED(DISABLE_Z_MIN_PROBE_ENDSTOP)) // It's possible for someone to set a pin for the Z Probe, but not enable it.
+        #if ENABLED(Z_MIN_PROBE_REPEATABILITY_TEST)
+          #error You must have a Z_MIN or Z_PROBE endstop to enable Z_MIN_PROBE_REPEATABILITY_TEST.
         #else
-          #error ENABLE_AUTO_BED_LEVELING requires a Z_MIN or Z_PROBE endstop. Z_MIN_PIN or Z_PROBE_PIN must point to a valid hardware pin.
+          #error ENABLE_AUTO_BED_LEVELING requires a Z_MIN or Z_PROBE endstop. Z_MIN_PIN or Z_MIN_PROBE_PIN must point to a valid hardware pin.
         #endif
       #endif
     #endif
 
     /**
-     * Require a Z Probe Pin if Z_PROBE_ENDSTOP is enabled.
+     * Require a Z Probe Pin if Z_MIN_PROBE_ENDSTOP is enabled.
      */
-    #if ENABLED(Z_PROBE_ENDSTOP)
+    #if ENABLED(Z_MIN_PROBE_ENDSTOP)
       #if !PIN_EXISTS(Z_PROBE)
-        #error You must have a Z_PROBE_PIN defined in your pins_XXXX.h file if you enable Z_PROBE_ENDSTOP.
+        #error You must have a Z_MIN_PROBE_PIN defined in your pins_XXXX.h file if you enable Z_MIN_PROBE_ENDSTOP.
       #endif
-      #if Z_PROBE_PIN == -1
-        #error You must set Z_PROBE_PIN to a valid pin if you enable Z_PROBE_ENDSTOP.
+      #if Z_MIN_PROBE_PIN == -1
+        #error You must set Z_MIN_PROBE_PIN to a valid pin if you enable Z_MIN_PROBE_ENDSTOP.
       #endif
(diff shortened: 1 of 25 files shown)


clefranc:
@AnHardt Right now, only these boards have a `Z_MIN_PROBE_PIN`:

**pins_AZTEEG_X3_PRO.h
pins_RAMPS_13.h
pins_RIGIDBOARD.h**

I'm using **pins_RUMBA.h**, and added the define manually because it was not there.

A lot of other boards already have a `Z_MIN_PIN` defined, can you update them so they also include the `Z_MIN_PROBE_PIN`?

AnHardt:
@clefranc 
Definitely not. I'd have to know a free usable pin for every board. I have and know only RAMPS-boards. Reading 20+ manuals is a bit to much.
```
#define Z_PROBE_PIN -1
```
or even worse
```
#if defined(Z_PROBE_ENDSTOP)
  // Define a pin to use as the signal pin on Arduino for the Z_PROBE endstop.
  #define Z_PROBE_PIN -1
#endif
```
does not make much sense to me.

I'd prefer the probe_extra_pin would go away at all. I really can't suggest what homing to z-min-endstop, but leveling with a probe is good for. It accomplishes to have the same height after G28 and G29.

Marlin/SanityCheck.h
```diff
@@ -133,37 +133,37 @@
      * Require a Z Min pin
      */
     #if Z_MIN_PIN == -1
-      #if Z_PROBE_PIN == -1 || (DISABLED(Z_PROBE_ENDSTOP) || ENABLED(DISABLE_Z_PROBE_ENDSTOP)) // It's possible for someone to set a pin for the Z Probe, but not enable it.
-        #if ENABLED(Z_PROBE_REPEATABILITY_TEST)
-          #error You must have a Z_MIN or Z_PROBE endstop to enable Z_PROBE_REPEATABILITY_TEST.
+      #if Z_MIN_PROBE_PIN == -1 || (DISABLED(Z_MIN_PROBE_ENDSTOP) || ENABLED(DISABLE_Z_MIN_PROBE_ENDSTOP)) // It's possible for someone to set a pin for the Z Probe, but not enable it.
```
thinkyhead: #if !PIN_EXISTS(Z_MIN_PROBE) || DISABLED(Z_MIN_PROBE_ENDSTOP) || ENABLED(DISABLE_Z_MIN_PROBE_ENDSTOP)

Marlin/SanityCheck.h
```diff
@@ -133,37 +133,37 @@
      * Require a Z Min pin
      */
     #if Z_MIN_PIN == -1
```
thinkyhead: I think this (and others like it in this file) should probably be: `#if !PIN_EXISTS(Z_MIN)` because leaving out the definition completely should also invoke these checks.

Marlin/SanityCheck.h
thinkyhead: #if !PIN_EXISTS(Z_MIN_PROBE) || DISABLED(Z_MIN_PROBE_ENDSTOP) || ENABLED(DISABLE_Z_MIN_PROBE_ENDSTOP) // It's possible for someone to set a pin for the Z Probe, but not enable it.

I'm not sure the comment on this line is correct, either.

Wackerbarth: @thinkyhead -- If the comment isn't correct, it should be.  In general, the pin file describes the capabilities (and specifies the addressing parameters for each).
Optional capabilities may, or may not be utilized. But that choice should not affect the definition of the board pins.

Marlin/stepper.cpp
```diff
@@ -411,7 +411,7 @@ inline void update_endstops() {
           #endif // !Z_DUAL_ENDSTOPS
         #endif // Z_MIN_PIN
 
-        #if ENABLED(Z_PROBE_ENDSTOP)
+        #if ENABLED(Z_MIN_PROBE_ENDSTOP)
```
thinkyhead: Is the fix suggested by @clefranc still needed here, or does `Z_MIN_PROBE_ENDSTOP` now cover this?
```cpp
#if ENABLED(Z_MIN_PROBE_ENDSTOP) && DISABLED(DELTA)
```

Marlin/SanityCheck.h
thinkyhead: The comment makes sense, actually, but I had to read it again. It explains why the 2nd and 3rd conditions apply.

thinkyhead:
Merging this in the hope that it addresses #61 and https://github.com/MarlinFirmware/Marlin/issues/2413

Marlin/stepper.cpp
clefranc: @thinkyhead My fix was for the `HAS_Z_MAX` block, @AnHardt removed its original `#if ENABLED(Z_PROBE_ENDSTOP)` block. For me and my delta, it's good, but with this merge, for cartesian users who have a Z max endstop only AND use `Z_MIN_PROBE_ENDSTOP`, the probe endstop is now ignored, ie crash...

PR #66, opened 2015-08-10
Remove the remaining ENABLED() and DISABLED() macros from the pin-files

AnHardt:
by replacing them with defined() and !defined().


Marlin/pins_MEGATRONICS_3.h
```diff
@@ -9,7 +9,7 @@
 #define LARGE_FLASH        true
 
 
-#if ENABLED(Z_PROBE_SLED)
+#if defined(Z_PROBE_SLED)
   #define SLED_PIN         -1
```
Wackerbarth: Why is this (any any others) be guarded at all?  Shouldn't it be the pin that will be used if the sled feature is to be used? (Admittedly, here, there is no pin assigned and that should cause a Sanity error if you try to enable the sled)

So far, the only potential use that I see for a guard relates to the fact that the commonly used SD card library already defines certain pins related to SPI.
But in that case, I believe that the compiler is quite OK with two #defines as long as they are the same.

If the intention is to assign an alternate set of pins for a second SPI channel, then we should not be doing that by overloading the bottom level definition of the pins, but by actively selecting an alternative.

Marlin/pins_PRINTRBOARD.h
```diff
@@ -38,14 +38,14 @@
...
 #else
   #define FAN_PIN          16
 #endif
 
 #define X_STOP_PIN         35
-#if ENABLED(SDSUPPORT)
+#if defined(SDSUPPORT)
   #define Y_STOP_PIN       37 // Move Ystop to Estop socket
 #else
   #define Y_STOP_PIN        8 // Ystop in Ystop socket
```
Wackerbarth: This is an optional wiring harness choice made by the user. Therefore "ENABLED" is appropriate.

AnHardt:
Feel free to eliminate all these conditions.
Here i do not try to solve a theoretical but a practical problem.
At lest i haven't seen any other attempt to resolve https://github.com/MarlinFirmware/Marlin/issues/2581.

AnHardt:
Sorry nonsense.

PR #70, opened 2015-08-11
Suppress some compiler warnings

Wackerbarth:
Cleanup -- No functional change.  But it gets rid of SOME of the compiler complaints about unused variables, etc.


thinkyhead:
Oops! `configuration_store.cpp:171: error: 'uint' was not declared in this scope`. It should probably be `uint8_t`.

PR #74, opened 2015-08-11
Add final Z raise after Delta probe

thinkyhead:
Do a final raise on Deltas that don’t already put away an allen key probe.
- This may only be patching an issue where the last probe point is culled from the points-to-probe. In that case, we should set a flag to that effect, or look at the last probe-action actually sent to `probe_pt()` and only raise if the last probe action didn’t include `Stow` – which would cause `probe_pt` to have already included the raise.
- Alternatively, the probe point loop can remember which index was the last one stored, and send the `Stow` action for that point.
- The final probe point problem I’m trying to address here may actually be fixed by #63 if it already remembers which probe point will be last.


Wackerbarth:
Why not just replace the "STOW" action with the "raise if needed" action?

thinkyhead:
I'm not sure how this PR currently applies, and I forget which comment / issue it was meant to address. I believe this applies only to deltas that probe using the nozzle, pushing down on the bed to trigger the switch. For these probes there's no stowage, and I believe the issue was that the nozzle was just left down on the bed at the end of `G29`. Testers with deltas and various probes should see if the current code is doing the right stuff, or if a fix similar to this is needed.

Wackerbarth:
The checks are failing ----

thinkyhead:
Ready for evaluation. Is such a raise needed at the end? I haven't been following testing of bed leveling in the RC branch. If delta users are doing ok, this patch might be unnecessary.

thinkyhead:
Seems to already be merged.

PR #75, opened 2015-08-11
Simplified Z_RAISE options

thinkyhead:
- Cleaned up redo of #62

Original notes, updated:

- Rename `Z_RAISE_BEFORE_PROBING` and `Z_RAISE_AFTER_PROBING` to `Z_PROBE_CLEARANCE` because they serve the same purpose.
- Rename `Z_RAISE_BEFORE_HOMING` and `Z_RAISE_BETWEEN_PROBINGS` to `Z_RAISE_FOR_OBSTACLES` because they serve the same purpose.
- Kill a small bug in `raise_z_for_servo()` where `Z_RAISE_BEFORE_HOMING` was used instead of `Z_RAISE_BEFORE_PROBING`.


boelle:
i can wote for this one...

@AnHardt is german and some stuff does not translate so well

nobody is a crazy fool... i assume that the intentions behind is ok, but yeah translations can make it look... lets say less ok

Wackerbarth:
@boelle - "nobody is a crazy fool" -- I will give Scott the benefit of believing that he did not mean the term in a derogatory manner. What can I say? I certainly present some "Wacky" ideas :)

As you note, the originator's native language is not the language in which the label appears to be written. The labels are somewhat arbitrary (To the compiler, Z47PZQ would work just as well).

So an improved "translation" might be helpful to those of us for whom English is our primary language.

thinkyhead:
> that crazy fool (as opposed to _this_ crazy fool)
> nobody is a crazy fool

I believe we are all crazy fools, but I can only authoritatively assert my own crazy foolishness.

thinkyhead:
I have some additional thoughts on this PR…

Since we're using a "clearance" concept instead of an "additional lift" concept, and since this will apply to all probes and not just to servo probes, then I think there ought to be a function similar to `raise_z_for_servo` (which may replace that function) which moves Z directly to the necessary height, based on the probe's lowest point (or based on the nozzle —aka Z=0— if it is lower). This will move Z up to the necessary height on min-homing or down on max-homing, then deploy or stow the probe as-needed. This function can then be called at any time the probe needs to be deployed or stowed, regardless of the current Z, and then it can just be called at the top of `deploy_z_probe`/`stow_z_probe`. I'd also like it if we unify the functions that deploy/stow probes of all kinds so there is a single call to deploy/stow all forms of probes, which manages all the raising logic.

Wackerbarth:
@thinkyhead - I certainly agree with the approach that you describe in your last comment. I have long felt that there should be a lot more uniformity in the handling of probing where the peculiar operations for sleds. Allen keys, and servos are isolated to small areas of the code. A sensor that does not require stowing can still go through that step in the procedure. It will simply have an empty set of actions needed to complete the step.

I am a bit confused about probing and max_endstops. Isn't probing always toward the minimum? If so, I fail to see how the maximum stop comes into play.

thinkyhead:
> I am a bit confused about probing and max_endstops. Isn't probing always toward the minimum? If so, I fail to see how the maximum stop comes into play.

Well, it's jut one of the available options, even if it's rare. You can use a max endstop for homing (as on a Delta), but then use a probe only for leveling. The main advantage of a Z max endstop (on a cartesian) is that wouldn't don't ever have to move it (as you might when changing extruders, affecting nozzle height). You can just change the Z home position (or `home_offset[Z_AXIS]`). The obvious downside to a Z max endstop is that Z axes tend to be slooooooow.

Wackerbarth:
I understand some uses for the max z -- My question concerns the probing.  
"This will move Z up to the necessary height on min-homing or down on max-homing, then deploy or stow the probe as-needed."
Why would you move the nozzle down before deploying the probe?

thinkyhead:
> Why would you move the nozzle down before deploying the probe?

That will depend on the logic. The function can check that "Z is known and above the clearance height" and then not do the extra move. The only reason to have it move down first, deploy, and then continue, is that it will look kind of cool.

AnHardt:
Let's assume a min/max - servo probe. Stow position may be horizontal, -90° may be deployed as a max-probe and +90 as a min-probe. Here for the function as a max-probe it needs the same as for a min-probe. You need space to deploy - and there may be obstacles around the point you want to probe.

thinkyhead:
@AnHardt Wow, I didn't realize there was a way to have a single servo probe act as both a max endstop and a min endstop (or bed probe). The code isn't very explicit about that specific arrangement, but I guess it's implied that the same switch could be both a max endstop and a bed probe. Very interesting…

thinkyhead:
Rebased and ready to test!

thinkyhead:
Rebased, ready for testing and feedback. Some deltas seem to want a different raise value for before and after. But let's see how important that really is.

PR #76, opened 2015-08-12
Lowercase GCode support

thinkyhead:
Replaces https://github.com/MarlinFirmware/Marlin/pull/2396
Refers to issue: https://github.com/MarlinFirmware/Marlin/issues/2393

Though not part of the GCode standard, some forks have chosen to allow lowercase parameters. This patch allows lowercase GCode to be used anywhere at all.

Please submit your up or down vote for this change.


eboston:
I would vote no on this.

boelle:
also a big -1 one from me

we agreed to try and follow as much common standard as we can

just because a fork comes up with something does not mean we should invent a new standard... there is enough confusion allready

Wackerbarth:
I'll vote -0.25.  I don't feel that it really adds value and it does (marginally) decrease the ability to detect some "garbage in". However, I am not strongly against it. To me, it just fails the cost/value test.

thinkyhead:
I agree that this doesn't add any real benefit and would only promote ugly-looking GCode.

Issue #35, opened 2015-08-03
Two observed stepper signal discrepancies during diagonal moves


boelle:
Continuation of https://github.com/MarlinFirmware/Marlin/issues/2370

maicrodrop:
Small update:  Resetting TCNT1 to 0, if it's already passed OCR1A, works so far; been printing and testing with it.  Have also tested setting TCNT1 to (0.75*OCR1A) which works as well, though haven't tested it as thoroughly yet; this would be preferable, as it minimizes the dead time between ISR calls.

I've also been investigating the 1.5.6>1.5.7 compiler change further.  I added code to the ISR, to use the fastio.h code to write a pin HIGH at the beginning and LOW at the end of the ISR.  While this does change the absolute execution time of the ISR, compiling the same changed code in 1.5.6 and 1.5.7 at least gives a (theoretical) good relative comparison.  I found that an individual STEP pulse takes ~3.5µS in 1.5.5, and ~4.5µS in 1.5.7.  The entire ISR (measured between when the "ISR_signal" pin [defined in pins.h, so WRITE is used in the normal way] goes HIGH and then when it goes LOW) takes ~86µS in 1.5.5, and ~108µS in 1.5.7.  Still doesn't narrow down _why_, but at least shows that there is a definite execution time difference.

The four images below show the 'scope shots listed above; there are four-step-set captures compiled in 1.5.5 and 1.5.7, and single step zoom in both as well.  Yellow is Channel 1/X STEP; Blue is Channel2/Y STEP; Purple is Channel3/"ISR_signal", in these shots HIGH while the ISR is running (set HIGH when it starts, then set LOW at the very end).
![155_single](https://cloud.githubusercontent.com/assets/13168739/9093492/71747e18-3b7a-11e5-9975-4b16731f60ca.png)
![157_full](https://cloud.githubusercontent.com/assets/13168739/9093493/71833fb6-3b7a-11e5-9bbd-66391fe39d91.png)
![157_single](https://cloud.githubusercontent.com/assets/13168739/9093495/718c55d8-3b7a-11e5-9333-c06df7565edd.png)
![155_full](https://cloud.githubusercontent.com/assets/13168739/9093494/718a821c-3b7a-11e5-8bda-6ee181706081.png)


[...]

PR #81, opened 2015-08-16
FSR Sensor Support

thinkyhead:
Redo of #45 with more concise language and fewer options.

- The first commit simply sets up the basic option and pin definition.
- The second commit adds the ISR code to read the sensor

This PR is under discussion. Further commits will finish implementing the feature.


Wurstnase:
Don't forget ```#define PID_dT ((OVERSAMPLENR * 14.0)/(F_CPU / 64.0 / 256.0))```

thinkyhead:
Good catch @Wurstnase. We should figure out some way to automate that, basing it off the number of states in that routine, instead of having it just be a lonely number.

Wurstnase:
Also keep in mind that the temp isr is maybe not fast enough. At 80steps/mm and a speed of 10mm/s you only get one sample every 11 steps.

thinkyhead:
@Wurstnase Also, there's no special reason (that I know of) to only do one thing per ISR. The main reason it's structured the way it is (as I understand it) is to provide a delay between the Setup and Test steps. So perhaps instead of a list of discreet states it can just have two or three – Init, Setup, and Measure.

Wurstnase:
This is somehow I do it. But I'm not familiar with AVR.

On the Due I activated freerun. So any ADC is updated as fast as possible and I only take the values when I need them. So I can handle the slow ADC for the temperatures and for every step or more a value for the FSR.

The FSR could be then in the idle().

thinkyhead:
I'll have to peek at the code and see what other sensors are handled in the `idle()` routine (or the main program loop). I think thermocouples are updated in the main loop for their own reasons. My main concern is just to keep the ISR light, but I see very little overhead to any of this sensor reading, so unless it really turns out to bog things down, I see no problem updating all these sensors in the ISR — I doubt it would impact machine performance at all.

Wurstnase:
You want to update all ADC in one ISR-loop? I guess that this is to slow. On an Arduino Due the maximum speed is with a lot of tricks about 1MHz. 1MHz/14 ~ 70kHz. The AVR should be slower.

Wackerbarth:
I'm quite concerned that adding to the ISR burden is problematic. We are already getting complaints of motion and communications problems when attempting to run at high movement rates.

fmalpartida:
Most of those high rate movements are more associated to slow or errors in the serial unterface. With the sav or printrboard it shouldn't happen. I have confirmed it with the sav anyway.

Wackerbarth:
See /Marlin/issues/2525 and #35  -- They concern changes in the ISR re-trigger timing.

thinkyhead:
For the moment I'm leaving it alone — continuing to do only a single operation per temperature ISR. If some of these sensor tests can be moved to another point, such as `idle()`, we could try that.

With my latest commit the number of states (`TempState::StartupDelay`) is used in the `PID_dT` calculation, which means that the number of states can now be variable, so instead of always waiting for 14 firings it can wait for only as many loops as needed to read all the sensors.

CONSULitAS:
@thinkyhead 
The following blog post describes an interesting way to read analog pins in asynchronous ways:
http://cosa-arduino.blogspot.se/2013/03/the-analog-pin-classes-introduction-to.html
search for `sensor.sample_request();`

The idea is to start a sample in hardware and pull the result later, saving up to 1800 instruction cycles per sample.

It would be possible to start sample_requests for all pins, do something different (or wait) and then pull the results, without having longer delay than for reading a single pin.

What do you think?

And due to the fact that we have timing problems in ISR at high speeds perhaps the following is interesting, too:
http://cosa-arduino.blogspot.se/2013/03/benchmarking-pin-classes.html

I hope, that i will have more time in a few weeks to do some coding on my own again.

Jochen

fmalpartida:
I was going to suggest something similar but very much on those lines:
ISR (n-1): 
read ADC channels(n-2)
Configure conversion (n-1)

However, that would be great for just one ADC channel as they share converter with the rest. 

That would leave the option to have the ADC read back from an ISR with a convertion queue. ISR preemption shouldn't be a problem if we don't permit nested interrupts. 

Timer ISR triggers the first convention and the ADC ISR takes care of reprogramming reads. Acquisition delays shouldn't be a proble given the huge thermal inertia. This assumes that there is time left after processing Timer ISRs. 

The other thing is the averaging that takes place, I haven't had the chance to see if we have a low pass filter or if it makes sequential acquisition to average. To avoid load on ADC ISR convertion and averaging would be done on the read operation. 

Just sharing some thoughts from last night.

thinkyhead:
@CONSULitAS @fmalpartida Thanks for the helpful thoughts and links. I'll definitely check them out. I'm wondering whether this PR should be merged as-is (once rebased, of course), or with my proposed tweak to the `TempState` processing, _or_ whether I should look into the hardware-based ADC and implement that first. The only caveat I have with hardware-based ADC is that (as I understand it) we can't do that yet because of the architecture of Marlin. (There's another topic where I recall discussing this, and left it hanging because I'm not sure where to begin.)

thinkyhead:
One question… The comment on this line in the temperature ISR seems to imply that a certain amount of time will elapse between runs of the block, as if it was important:
```cpp
if (temp_count >= OVERSAMPLENR) { // 10 * 16 * 1/(16000000/64/256) = 164ms.
```
With more or fewer sensors the time value will of course differ, and could be anywhere from 32ms up to 230ms. So I'm wondering, is it important that there be some minimal amount of delay here? Or, is it only important that there be 16 (`OVERSAMPLENR`) loops between runs of this block?

Wurstnase:
> Or, is it only important that there be 16 (OVERSAMPLENR) loops between runs of this block?

This.

fmalpartida:
My personal impression is to get 1.1 out and then start with this change. 

Enviado desde mi iPhone

fmalpartida:
Here is you use the running ADC as I described, every time a conversion is finished you will get an interrupt. Conversion don't take that long, around de 120us mark on a 16MHz clock. If I remember correctly the datasheet specifies it in terms of clock cycles so on an 8MHz clock it should be double. In any case you also have an end of conversion flag in the HW. 
I'll try to took at the code and see how it matches with the HW. 

Enviado desde mi iPhone

thinkyhead:
@fmalpartida I agree that it will be good to put put 1.1 first and then add this sensor and tweak the measurement code. That will give us more time to go over this stuff and for me to get a better grasp on the hardware-level options available to us on AVR. (I'll post elsewhere on the subject of 1.1.0-RC1 so we can get the ball rolling on that.)

thinkyhead:
Rebased, in case anyone wants to go forward with testing and evaluation on this!

thinkyhead:
I've made a new PR #345 against `dev` for this feature to keep it alive. I will still need to explore using ADC for reading the sensors, but for now this patch should be about right to extend the current technique. Discuss further at #345.

PR #82, opened 2015-08-16
Fix Core_XZ width specific Nicholas Seward's

MagoKimbra:
Fix planner CORE_XZ width speific Nicholas Seward's
http://forums.reprap.org/read.php?2,377858


Wackerbarth:
This change is to support a particular kinematic configuration. If needs to be presented as a configuration option with conditional compilation as needed.

MagoKimbra:
Reading on the forum, I think I understand it must be so for CORE_XZ, and not setting a 1: 1 but rather 1: 3 between X and Z does not work otherwise. But I may have misread.

MagoKimbra:
I read this:
       X=x-3z
       Z=x+3z
       Y=y
       
      The capitals are steppers and the lowet case letters are coordinates.
      
     I would be interested in a sketch of what you have planned. I can't visualize it yet.

Wackerbarth:
You are correct about 3:1.  The issue that I am raising is that only some CORE_XZ printers have this configuration. The Marlin code needs to work on ALL of the designs. By creating an option in the configuration file, we can select, at compile time, which mechanical version we have.

So, in the configuration we have
// #define CORE_XZ
// #define CORE_X3Z      // This enables Nich...   block_and_tackle design ...
...

And, in the code, we will conditionally compile appropriate code for whichever option was selected

MagoKimbra:
OK, I apologize for this, I thought it was a thing of default for CORE_XZ.
Then I close?

Wackerbarth:
No, don't close it. Make it an option.

MagoKimbra:
A ok!!

PR #83, opened 2015-08-16
Don't free stack-based plane_equation_coefficients

thinkyhead:


Wackerbarth:
Perhaps a slight rewording of the explanation --
The plane_equation_coefficients are no longer allocated from the heap. Therefore they don't get released by "free". Rather, as part of the stack, they "release" automatically as the routine goes out of scope.

PR #84, opened 2015-08-16
Singlenozzle

MagoKimbra:
In configuration.h add
//#define SINGLENOZZLE
When enable this option only hotend is set for more extruder...


Wackerbarth:
You need to separate the CoreXZ commit from this submission. They are independent.
Also, I don't really understand just what this is.
I'm guessing that you have one nozzle that is fed multiple filament strands so that you can extrude a multi-color result. But I don't want to guess.
Please provide a better explanation.

A major part of your changes involving the renaming of EXTRUDER to HOTEND. That, in itself, does not change any logic.  So please separate your submission into two commits. The first would simply rename terms. (A global substitution of EXTRUDER for HOTEND, etc. would return us to the current code)

The second would have the logic changes.  That way, a reviewer and quickly scan the commit that alters many lines without affecting the logic and move on to looking at the actual logic changes.

MagoKimbra:
I closed and do it new...

Wackerbarth:
OK. But you could have left it open and simply committed (perhaps with `git push -f`) a new version of your SINGLENOZZLE branch.

PR #85, opened 2015-08-16
COREX3Z

MagoKimbra:
Add system COREXZ by Nicholas Seward's.
http://forums.reprap.org/read.php?2,377858
This system have this coordinate:
X = X - 3Z
Z = X +3Z
Y = Y


AnHardt:
Next week we'll have a request for a X5Z or a X1/3Z
How about a definable factor?
COREXZ_XZ_FACTOR ?
Obviously also for CoreXY.
COREXY_XY_FACTOR .

It than also could be negative by the users will.

Wackerbarth:
Actually, We should be able to unify Cartesian, Core_XY, and Core_XZ by defining a few ∂A/∂X, ∂B/∂X, ... constants which plug into the general matrix transform.  I assume that the compiler would be smart enough to optimize `da = 1*dx + 0*dy + 0*dz;` into `da = dx;`

MagoKimbra:
I also I was thinking about a solution like this, a multiplier to enter as a value. And leave everything as it was before. I'm trying various solutions just can edit. In fact this solution that I put I do not like ...

Wackerbarth:
The place where I was suggesting putting this has already converted to integer. If will work fine for unification of Cartesian, CORE_XY and CORE_XZ because all of the coefficients as in {-1, 0, 1}

I'm thinking that you are looking at X3Z incorrectly. Can't you simply handle the "3" as a scaling in the conversion of Z from mm to steps?  Once it is built, it is still just a CORE_XZ mechanism.

So, my conclusion is that we don't need this at all. You simply adjust the conversion coefficients.

It would be like replacing a 30 tooth drive pulley with a 10 tooth pulley in a Cartesian system.

PR #86, opened 2015-08-16
SINGLENOZZLE

MagoKimbra:
Add option for SINGLENOZZLE.

How does it work:
If you have more extruders, but one hotend, see the Cyclops E3D or similar type FluxCapacitor, you can set SINGLENOZZLE so the FW see only a hotend. All calls to the single temperature ranging hotend.
Then the FW defines a single port for the sensor and also that of power for the resistance.

In conditional.h it is:
```cpp
#if ENABLED (SINGLENOZZLE)
  #define HOTENDS 1
  #undef TEMP_SENSOR_1_AS_REDUNDANT
#else
  #define HOTENDS EXTRUDERS
#endif
```
So if that is not defined SINGLENOZZLE everything returns to normal HOTENDS = EXTRUDERS, then filling is replaced with the same number in all files.
But if that is defined SINGLENOZZLE then HOTENDS = 1.
It is called a single port sensor and a single power port.
Calls M104 and M109 are always at the same sensor.
in temperature.h
```cpp
#if HOTENDS <= 1
  #define HOTEND_ARG 0
#else
  #define HOTEND_ARG hotend
#endif
```
This system has been operating for more than a year of many printers with FluxCapacitor. http://www.thingiverse.com/thing:498922
It works perfectly and allows the change temeprature underway for multi material, this is because only the commands M104 Txx are executed if the extruder is the same as the current command.

Example:
Extruder current T1 to exchange in T3
In gcode change switches is
```gcode
M104 T0 S190
M104 T1 S195
T2 M104 S200
T3 M104 S260
```
Only the command M104 S260 T3 runs and set only hotend to 260.


Marlin/Configuration.h
```diff
@@ -85,11 +85,15 @@ Here are some standard links for getting your machine calibrated:
 // :[1,2,3,4]
 #define EXTRUDERS 1
 
+// This is used for single nozzle and multiple extrusion configuration
+// Uncomment below to enable (One Hotend)
```
Wackerbarth: All in all, the PR looks very reasonable.

But, perhaps, we should have a more explicit description, such as "For Cyclops E3D and similar type extruder configurations with multiple feeds sharing a single nozzle and heater."

@thinkyhead -- Any other "style" issues?

thinkyhead:
It looks very good to me. The only thing I would add is a test in `SanityCheck.h` for the old settings names, throwing an error to tell users the names have been changed. I've just been adding more `#elif` lines at the end for this kind of thing.

boelle:

jbrazio:

PR #91, opened 2015-08-18
Fix Make Z_PROBE a Z_MIN_PROBE #65

clefranc:
* Fix a couple of renames omission and macro expansion errors.


clefranc:
@Wackerbarth Are you going to produce a new PR to remove the _MIN, do you want I try to produce this new renaming PR, or the probe rename remains?

Wackerbarth:
@clefranc -- I would be happy to have you produce a PR of this nature.  Overall, I'm looking to see a unification of the various probing schemes so that they can share more code and we can reduce some of the #ifdef clutter that makes the present code extremely difficult to maintain.

In that direction, I think of a PROBE which has a PROBE_SENSOR_PIN and, perhaps, a PROBE_DEPLOYMENT_PIN (to drive a magnet that holds the sled, etc.)

Unlike the earlier delta code that called the probe a MIN switch, I consider a MIN sensor, such as a FSR, to be an alternate implementation of a PROBE.

I merged the Z_MIN_PROBE changes because the existing code had obvious errors and the PR at least reduced them. But a better cleanup would be welcome.

clefranc:
@Wackerbarth Is BED_LEVEL_SENSOR a good alternative? Trying to merge the "conventional" probe and FSR term to their actual purpose.

```
Z_MIN_PROBE_ENDSTOP = BED_LEVEL_SENSOR
ENDSTOPPULLUP_ZMIN_PROBE = BED_LEVEL_SENSOR_PULLUP
Z_MIN_PROBE_ENDSTOP_INVERTING = BED_LEVEL_SENSOR_INVERTING
DISABLE_Z_MIN_PROBE_ENDSTOP = DISABLE_BED_LEVEL_SENSOR
```
What do you think?

Wackerbarth:
BED_LEVEL_SENSOR == What does the probe have to do with "level"? (Yes, we can use it to assist in determining how to level the bed, but the probe only senses contact with the surface below the extruder, level,or not)  It's just a probe, perhaps I want to use it to measure the height of objects that I place on the bed. Then it isn't even measuring to the bed.

clefranc:
@Wackerbarth I understand.

Natealus:
And if one wanted to use FSR on their delta they could possibly be placed as XYZ mins under each tower and if some clever coding was created, we might be able to tell when a part was being pushed on unevenly and provide warning of miscalibration or maybe even adjust in real time. That's probably a very tricky bit of coding but many things are possible. :)

Unfortunately from what I've heard FSR don't work well under heat so monitoring during ABS prints are for now out of the common category.

PR #93, opened 2015-08-19
update for newest version of U8glib_Arduino

iplayfast:


oysteinkrog:
This needs to be rebased, currenly it is a mess.

thinkyhead:
Is this backward-compatible for the older Arduino toolset? Does it matter to us going forward?

Wackerbarth:
@thinkyhead -- This change replaces the direct inclusion of the c interface file with the inclusion of the c++ interface file. That file performs the above mentioned inclusion. Therefore, it should work on any compilation. Further, the loader should drop the c++ wrapper since it is not called.

Rebased version appears as the pr/93 branch

Wackerbarth:
Merged

PR #94, opened 2015-08-20
G5 X Y Z Babystepping

gralco:
G5 X Y Z — The argument is the number of steps to be performed in the positive or negative direction of the given axes. It attempts to prevent endstop crashes to the best of its ability. *** Added the option to babystep the stored `home_offset[Z_AXIS]` if `BABYSTEP_OFFSET` is defined (this can also be performed on an LCD controller, and saved to the EEPROM as with standard home-offsets) ***

### DIFF
--- a/Marlin/Marlin_main.cpp
+++ b/Marlin/Marlin_main.cpp
@@ -85,6 +85,7 @@
  * G2  - CW ARC
  * G3  - CCW ARC
  * G4  - Dwell S<seconds> or P<milliseconds>
+ * G5  - Babystep Movement X Y Z
  * G10 - retract filament according to settings of M207
  * G11 - retract recover filament according to settings of M208
  * G28 - Home one or more axes
@@ -1104,9 +1105,16 @@ static void set_axis_is_at_home(AxisEnum axis) {
     else
   #endif
   {
-    current_position[axis] = base_home_pos(axis) + home_offset[axis];
-    min_pos[axis] = base_min_pos(axis) + home_offset[axis];
-    max_pos[axis] = base_max_pos(axis) + home_offset[axis];
+    if(axis != Z_AXIS)
+      current_position[axis] = base_home_pos(axis) + home_offset[axis];
+    else
+      current_position[axis] = base_home_pos(axis)
+      #ifndef BABYSTEP_OFFSET
+        + home_offset[axis];
+        min_pos[axis] = base_min_pos(axis) + home_offset[axis];
+        max_pos[axis] = base_max_pos(axis) + home_offset[axis]
+      #endif
+      ;
 
     #if ENABLED(AUTO_BED_LEVELING_FEATURE) && Z_HOME_DIR < 0
       if (axis == Z_AXIS) current_position[Z_AXIS] -= zprobe_zoffset;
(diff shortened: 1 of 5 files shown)


Wackerbarth:
A question about the design of this command:  Why do you perform this operation in some sequence of the axes rather than "all at once"? If someone wants to do the moves sequentially, they could issue a sequence of G5 commands. Thus as you describe G5 X Y Z, it would be equivalent to G5 Z, G5 X, G5 Y.

Also, if you have a good reason to avoid the above, shouldn't the order be determined by the order on the command line?  That way the user would have control over the order.

gralco:
@Wackerbarth 
I'm sorry, my description in the initial comment is not indicative of the implementation. Each one is actually performed *simultaneously*. The order seems to be arbitrary now, I thought it would make the code more elegant. I initially did this because the user can enable BABYSTEPPING but disable BABYSTEP_XY. 

Your comment makes complete sense and I'm going to push a commit right now that puts Z at the end of the array. Thanks for the input!

Marlin/Marlin_main.cpp
```diff
@@ -1794,6 +1802,26 @@ static void homeaxis(AxisEnum axis) {
       #endif
     }
 
+    #if ENABLED(BABYSTEPPING)
+      if(axis == Z_AXIS)
+      {
```
thinkyhead: Sorry to nag on minor points of style, but two things: According to our [style guidelines](http://www.marlinfirmware.org/index.php/Coding_Standards), it should have a space after `if`, and the opening brace should be at the end of the line.

Marlin/Marlin_main.cpp
```diff
@@ -2229,6 +2320,16 @@ inline void gcode_G28() {
 
     sync_plan_position();
 
+    if(code_seen(axis_codes[Z_AXIS])) {
+      if(code_value_long() != 0) {
+        current_position[Z_AXIS]=code_value()
```
thinkyhead: Similarly, in most cases operators such as `=` should be padded with spaces.

Marlin/Marlin_main.cpp
gralco: @thinkyhead 
No problem, some accidentally slipped. I just pushed a commit which should fixes them, please let me know if you see any more you want fixed.

Marlin/Marlin_main.cpp
```diff
@@ -2013,6 +2021,23 @@ static void homeaxis(AxisEnum axis) {
       #endif
     }
 
+    #if ENABLED(BABYSTEPPING)
+      if (axis == Z_AXIS) {
+        baby_max_endstop[axis] = Z_BABY_DEFAULT_MAX_POS;
+        baby_min_endstop[axis] = Z_BABY_DEFAULT_MIN_POS;
+      }
+      #ifdef BABYSTEP_XY
```
thinkyhead: This should be `#if ENABLED(BABYSTEP_XY)` as with other switches.

Marlin/Marlin_main.cpp
```diff
@@ -1104,9 +1105,16 @@ static void set_axis_is_at_home(AxisEnum axis) {
...
   #endif
   {
-    current_position[axis] = base_home_pos(axis) + home_offset[axis];
-    min_pos[axis] = base_min_pos(axis) + home_offset[axis];
-    max_pos[axis] = base_max_pos(axis) + home_offset[axis];
+    if (axis != Z_AXIS)
+      current_position[axis] = base_home_pos(axis) + home_offset[axis];
+    else
+      current_position[axis] = base_home_pos(axis)
+      #ifndef BABYSTEP_OFFSET
```
thinkyhead: This should be `#if DISABLED(BABYSTEP_OFFSET)` as with other switches.

boelle:
could this PR be rebased against RCbugfix on the main repro? we are going to cut this repro loose as it just adds to the confusioncould this PR be rebased against RCbugfix on the main repro? we are going to cut this repro loose as it just adds to the confusion

PR #97, opened 2015-08-20
Set Extruder Offsets on an LCD Controller and Save to EEPROM

gralco:
Title says it all.

### DIFF
--- a/Marlin/Configuration.h
+++ b/Marlin/Configuration.h
@@ -88,8 +88,8 @@ Here are some standard links for getting your machine calibrated:
 // Offset of the extruders (uncomment if using more than one and relying on firmware to position when changing).
 // The offset has to be X=0, Y=0 for the extruder 0 hotend (default extruder).
 // For the other hotends it is their distance from the extruder 0 hotend.
-//#define EXTRUDER_OFFSET_X {0.0, 20.00} // (in mm) for each extruder, offset of the hotend on the X axis
-//#define EXTRUDER_OFFSET_Y {0.0, 5.00}  // (in mm) for each extruder, offset of the hotend on the Y axis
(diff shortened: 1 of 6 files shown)


oysteinkrog:
Please fix the commit message to describe the changes.

thinkyhead:
This is a subtle thing, but it's a good idea to always write the same number of values out to EEPROM even if the number of extruders differs. That way the EEPROM values always align, even if yuo re-flash with more or fewer extruders. So in this case, it should always write out and read 4 extruder offsets. When writing them, just use zeros for the unused ones. When reading them, you shouldn't need to to anything special, just read all 4. Also, when making any change to EEPROM, always bump the version number. I think the next one is "V21".

gralco:
@thinkyhead
That makes sense, so how's the recent commit?

thinkyhead:
@Wackerbarth Yes, in fact that's the idea. The global array only needs to be as long as necessary to store the offsets per extruder. The EEPROM write function then pads out the data using a temporary var, and the EEPROM read function reads the values into a dummy var before storing (or not storing) them to the global array.

Wackerbarth:
Agreed.But, unless I misread the proposed code, it actually allocates RAM to all possible variables.

Wackerbarth:
And along with that, we should pre-allocate space for 11 variables to describe the delta Space, (whichI think exceeds the number needed for CoreXX and Cartesian systems).

boelle:

studiodyne:
Hi ,
Can you make the added code to make a Configurator switch
To switch from chimera to cyclops hotend or diamond = LCDgralcomenu 'for equal offsets ' + ' Heater1/sensor1 equals Heater0/sensor0 ... to trick the sofware '

For the moment , if we want to switch between cyclops and chimera , we have to re'firm' the board
Can you make us , a Switch Menu to swap between our configuration or , just add  a LCD HEATER trick menu 
i'm sure i will be good in programmation in the future but for the moment , i need help 
thks

PR #99, opened 2015-08-20
M601 Resume The Print From Filament Change

gralco:
This nullifies the requirement for an LCD controller, therefore the preprocessor conditionals should be modified as well.

Specifically, Ultipanel does not have to be defined.

### DIFF
--- a/Marlin/Marlin_main.cpp
+++ b/Marlin/Marlin_main.cpp
@@ -195,6 +195,7 @@
  * M503 - Print the current settings (from memory not from EEPROM). Use S0 to leave off headings.
  * M540 - Use S[0|1] to enable or disable the stop SD card print on endstop hit (requires ABORT_ON_ENDSTOP_HIT_FEATURE_ENABLED)
  * M600 - Pause for filament change X[pos] Y[pos] Z[relative lift] E[initial retract] L[later retract distance for removal]
+ * M601 - Resume the print from filament change
  * M665 - Set delta configurations: L<diagonal rod> R<delta radius> S<segments/s>
  * M666 - Set delta endstop adjustment
  * M605 - Set dual x-carriage movement mode: S<mode> [ X<duplication x-offset> R<duplication temp offset> ]
@@ -420,9 +421,15 @@ bool target_direction;
   boolean chdkActive = false;
 #endif
 
+<<<<<<< HEAD
 #if ENABLED(PID_ADD_EXTRUSION_RATE)
   int lpq_len = 20;
 #endif
+=======
+#ifdef FILAMENTCHANGEENABLE
+  boolean change_filament = false;
+#endif //FILAMENTCHANGEENABLE
+>>>>>>> Added M601 Resume Print From Filament Change feature
 
 //===========================================================================
 //================================ Functions ================================
@@ -890,6 +897,10 @@ void get_command() {
 
       // If command was e-stop process now
       if (strcmp(command, "M112") == 0) kill(PSTR(MSG_KILLED));
+
+      #ifdef FILAMENTCHANGEENABLE
+        if (strcmp(command, "M601") == 0) change_filament = false;
+      #endif //FILAMENTCHANGEENABLE
 
       cmd_queue_index_w = (cmd_queue_index_w + 1) % BUFSIZE;
       commands_in_queue += 1;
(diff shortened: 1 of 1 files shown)


oysteinkrog:
Please fix the commit message to describe the changes.

Marlin/Marlin_main.cpp
```diff
@@ -4964,6 +4973,11 @@ inline void gcode_M503() {
     
   }
 
+/**
```
thinkyhead: The function and its header comment should be indented 2 spaces. I see that the `#endif` is indented 1 space, so it's probably off by a space, but you needn't worry about that. Just make sure indentation is as proper as you can make it based on the current code, with each level indented by 2 more spaces than the containing one.

Marlin/Marlin_main.cpp
```diff
@@ -876,6 +881,10 @@ void get_command() {
       // If command was e-stop process now
       if (strcmp(command, "M112") == 0) kill(PSTR(MSG_KILLED));
 
+      #ifdef FILAMENTCHANGEENABLE
+        if (strcmp(command, "M601") == 0) change_filament = false;
+      #endif //FILAMENTCHANGEENABLE
+
```
thinkyhead: Is this proved to be necessary? Every time we add a strcmp call like this it adds some overhead to the parser. If the buffer is being processed then the `gcode_M601` handler should be sufficient.

thinkyhead:
@gralco Eric, you're on a roll… good stuff!

Marlin/Marlin_main.cpp
gralco: @thinkyhead 
The buffer is not processed during the while loop here: https://github.com/gralco/MarlinDev/blob/M601-Exit-Change-Filament-Mode/Marlin/Marlin_main.cpp#L4917
The G-code is queued but not performed until the loop is exited. At least those were my results.
Really gcode_M601() serves essentially no purpose.

thinkyhead: I think you are correct. This appears to be the only way to break out of `M600` since Marlin currently sits in a program loop. And yes, it does make `gcode_M601` redundant. I suggest removing the code in `gcode_M601` completely and replacing it with a comment that the real implementation is earlier in the parser. Maybe later on we'll do something on an equivalent level, such as making a more generic gcode that causes `lcd_clicked` to return `true` – then we could use that code to break out of any code waiting for a click.

boelle:

thinkyhead:
> could this PR be rebased against RCbugfix on the main repro?

We have two ongoing filament change PRs currently in discussion together, coordinating to improve the auto and manual filament change features. They could both use more testing and encouragement to keep moving forward. MarlinFirmware/Marlin#3605 and MarlinFirmware/Marlin#3662

Side-note: You should say "repo" as in "**repo**sitory" (not "repro" as in "**repro**duction"). It will make you sound more git-savvy and geeky and stuff.

scalez:
@thinkyhead: @gralco is me, if `M601` is still useful to Marlin I can create a PR against [MarlinFirmware/Marlin](https://github.com/MarlinFirmware/Marlin). Since `M601` is handled in the parser like `M112` and `M108` then we would likely want to start using @AnHardt's`EMERGENCY_PARSER` macro for this feature as well.

scalez:
> Maybe later on we'll do something on an equivalent level, such as making a more generic gcode that causes lcd_clicked to return true – then we could use that g-code to break out of any loop waiting for a click.

@thinkyhead not sure if we still want to do this but the way @AnHardt's repo is using `M410` for breaking out of movements such as `G0`, `G1`, `G28`, `G29`, etc; and [`M108`](https://github.com/MarlinFirmware/Marlin/pull/3611) (as well as `M601` here) are each used to break out of wait-states.

Does it make sense to combine the functionality of `M601`, `M108`, and `M410` (and other possible wait commands)?

AnHardt:
Indeed the first idea for the emergency command parser was based on the idea to send only two single byte chars. CTRL_X to kill - like M112, and CTRL-C to break the current operation. CTRL chars are not used in g-code. Single chars are even faster to parse then 4 byte g-codes - would not need context checking. If we can convince the hosts to send them is an other question.

boelle:
please rebase this against the rcbugfix... this repro will go away

avluis:
@boelle It's repo btw. Just an FYI.

PR #100, opened 2015-08-20
Added support for SSD1306 and SH1106 based OLED LCD displays.

fmalpartida:
Added support for OLED LCDs based on SSD1306 and SH1106 drivers.
Previous pull requested added to 1.0.3 release.


Marlin/Conditionals.h
```diff
@@ -44,6 +44,13 @@
   #if ENABLED(U8GLIB_SSD1306)
```
thinkyhead: Looks like this could be…
```cpp
#if ENABLED(U8GLIB_SH1106) || ENABLED(U8GLIB_SSD1306)
    #define ULTRA_LCD  //general LCD support, also 16x2
    #define DOGLCD  // Support for I2C LCD 128x64 (Controller SSD1306 graphic Display Family)
    #define ULTIMAKERCONTROLLER
   #endif
```

fmalpartida: Good catch.

thinkyhead:
Darn, needs another rebase. I'll wait for that before I try to review the code again.

fmalpartida:
Rebased and updated.

Wackerbarth:
You performed a merge, not a rebase. This needs to be rebased.  It should not contain ANY parts of the commits that others have made.

fmalpartida:
Brrrrr! How do you rebase it? Just click on the rebase?

Wackerbarth:
As bad as things have gotten, rather than rebase, I suggest that you effectively "start over".
Make a copy of the files that you have actually changed. Then check out the current "master" and replace those files with your copy. Then look at the differences created and keep the changes that you introduced.

But remember that you need to also make the same changes to EVERY example Configuration.

fmalpartida:
Mutiny...

fmalpartida:
@Wackerbarth did you pull in your development branch changes to the main code? I am about to go into a round of testing and had to re-merge in my configuration branch the changes for testing.

thinkyhead:
@fmalpartida Hey Francisco, sorry to hijack this thread. I heard from Richard that there may have been some regressions due to recent changes in the LCD display code, but I don't know the specifics.

The changes I proposed in #161 are all cosmetic, but I might have missed a subtle bug.

Pull request #165 was posted for testing, but it should have been kept on hold because it has the side-effect of making "Card Inserted" appear on the display if a card is already in the slot on startup. The workaround for this is to make `lcd_sd_status` into a `uint8_t` instead of a `bool` and make a distinction between "no card in slot" and "unknown." If the status is "unknown" before the SD pin is tested then the message can be suppressed.

fmalpartida:
No worries…

I don’t think that that particular part of the code has some regressions.

What I can confirm is that the current development branch is working very nicely.

What it is bugging me a bit about this PR is the fact that I has gone through and hopping between Marlin and MarlinDev not once, not
not three-ice, four times! 

First time I made the changes in Marlin/dev branch we were carrying forward, it didn’t get pulled in so, then it went into MarlinDev a good
range of the code got changed in the interim and when it was time to get it pulled in it couldn’t as there were conflicts. Messed up the rebase
(well, merged instead of rebasing) and finally it went on Richard’s configuration experimental branch.

So right now, I have given up on it and closed it waiting for better times.

Wackerbarth:
> On Sep 14, 2015, at 3:40 PM, fmalpartida <notifications@github.com> wrote:
> What it is bugging me a bit about this PR is the fact

I guess that I never figured out just what you were doing with SAV_3DGLCD.

Generally, it looks to be nothing more than a convenience declaration which can equally be specified by a couple of our existing selectors.

I find it hard to justify adding yet another way to skin the same cat.

Also, your changes were provided to the Configuration.h file without the corresponding changes to all of the other copies of it.

Putting it in my experimental branch eliminated those issues. But, like yourself, I have grown tired of attempting to chase the ever changing configuration mess.

fmalpartida:
Its a brand new OLED LCD which currently is working just fine.

The part on the Configuration file is a declaration of the LCD, I don’t think that right now there is support for any
OLED LCDs based on the SSD1306 and SH1106 controller chips which is what the PR brings forward.

The way I’ve packaged the PR is to enable support for the SSD1306 and SH1106 so that anyone can easily build up
an LCD module with what is available. I should have been more sneaky ( :-) ) and added it like the rest of the LCD configurations
where most of them are a convenient remix of the same smart controller.

I am very much with you on the current state of the configuration, hence my call to mutiny! Right now is a great mess which is causing
all sorts of trouble and unnecessary overhead.

PR #108, opened 2015-08-28
Clean up Dual Y and Dual Z

thinkyhead:
This change allows Dual Y and Dual Z support to be enabled for any board that has a second (unused) set of extruder stepper pins (E1). It also removes the assumption that Dual Z steppers will use E2 (rather than E1).

This is a more comprehensive solution than that proposed by @d00616 in https://github.com/MarlinFirmware/Marlin/issues/1784


Wackerbarth:
" Finally, it updates the example config for Kossel Pro, which was quite out-of-date."

Doesn't this indicate e that there is a fundamental error in the way that we are describing configurations?
A Kessel Pro is a DELTA configuration. I cannot see how how any "XY" or "XZ" definitions should affect that configuration.

thinkyhead:
@Wackerbarth The update to the Kossel Pro config is incidental. Somehow it either never got updated or it got reverted to some older version. After the commit included here which patches it, it barely differs from the Kossel Mini config. That commit could be pulled out and placed into a separate PR, but I happened to include it here because all the other configs are getting patched.

Wackerbarth:
"The update to the Kossel Pro config is incidental. Somehow it either never got updated or ..."

which just exemplifies my contention that we have a SERIOUS problem in the way that we handle configuration.

thinkyhead:
@Wackerbarth You're preaching to the choir. My pat reply remains: please submit a PR implementing a better configuration system.

Wackerbarth:
I feel that I have done so. However, YOU have rejected all such suggestions because they change the idea on "EVERYONE must have EVERYTHING presented in the ONE file that "matches" the (de jour) default configuration.

Wackerbarth:
BTW, I rebased this for you last week. Adopt the MarlinDev/pr/108 branch as a replacement to eliminate most of the extraneous changes that get pulled in because your patch was based on an older branch point. (There are probably some additional changes that have occurred more recently. So there may still be some conflicts. But the number of them should be reduced.)

thinkyhead:
> YOU have rejected all such suggestions

I'm not the only developer here, and I sure don't feel like I have any power to reject anything (whether a PR or an idea) to the point where you should abandon it altogether. Others with merge privileges may instantly recognize the value of your awesome PR and merge it. You should keep it alive for wider consideration.

PR #112, opened 2015-08-30
Easy Bed Leveling Rectangle

IamTheVector:
This edit is made to solve the  awkward method actually used to find the right probing square. This edit let the user make use of the X and Y value printed on screen on the GUI and find the lower leftmost point and the rightmost top point of the square in wich to probe. The process is simple. 

-Home Axis and rise the Z to let the space for the probe
-Extend the probe
-Bring the probe where you want the leftmost front point. Use m114 to print the actual position write down X and Y value. 
-Write X and Y value in the POS_A_X and POS_A_Y 
-Bring the probe where you want the rightmost back point. Use m114 to print the actual position write down X and Y value. 
-Write X and Y value in the POS_B_X and POS_B_Y 
- Still , every value has  to be consistent. You cant probe a point that your probe can't reach.  
End. ( Explaination video incoming )

### DIFF
--- a/Marlin/Configuration.h
+++ b/Marlin/Configuration.h
@@ -465,10 +465,36 @@ const bool Z_MIN_PROBE_ENDSTOP_INVERTING = false; // set to true to invert the l
 
   #if ENABLED(AUTO_BED_LEVELING_GRID)
 
-    #define LEFT_PROBE_BED_POSITION 15
-    #define RIGHT_PROBE_BED_POSITION 170
-    #define FRONT_PROBE_BED_POSITION 20
-    #define BACK_PROBE_BED_POSITION 170
+/*
+                   BACK
+         |||||||||||||||||||Y||||||||
+         |||||||||||||||||||Y||||||||
+         ||||||||----------(B)XXXXXXX
+         ||||||||           |||||||||
+ LEFT    |||||||| Probing   ||||||||| RIGHT
+         |||||||| rectangle |||||||||
+         XXXXX(A)-----------|||||||||
+         ||||||Y|||||||||||||||||||||
+         ||||||Y|||||||||||||||||||||
+                   FRONT
+Edit By Nader Al Khatib Fab Lab Frosinone officine giardino                                                                              
+*/
+
+
+// set the rectangle in which to probe
+#define POS_A_X 40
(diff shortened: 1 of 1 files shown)


Sniffle:
The problem with this configuration for decining the probe area is the
assumption that all home positions are in the front right.  This isnt
always the case as one of my printers home is in the back left.

IamTheVector:
Yeah thanks Sniffle, i think you are right. I just tought this solution, but i have to verify, but i think it should do the trick. 

 #define LEFT_PROBE_BED_POSITION  (POS_A_X-X_PROBE_OFFSET_FROM_EXTRUDER*X_HOME_DIR)
 #define FRONT_PROBE_BED_POSITION (POS_A_Y-Y_PROBE_OFFSET_FROM_EXTRUDER*Y_HOME_DIR)
 
 #define RIGHT_PROBE_BED_POSITION (POS_B_X-X_PROBE_OFFSET_FROM_EXTRUDER*X_HOME_DIR)
 #define BACK_PROBE_BED_POSITION  (POS_B_Y-Y_PROBE_OFFSET_FROM_EXTRUDER*Y_HOME_DIR)

IamTheVector:
How you achieve this configuration usually? By using MAX Home endstops, or just reversing the movements of you axis? 

Code talking:
#define X_HOME_DIR
or 
#define INVERT_X_DIR

Sniffle:
Moving the endstops and inverting the direction

IamTheVector:
I actually tried to solve the issue you found that was reasonable and understandable... i made math and sign inversion with if statement just to find out that it works with the same expression.

Sniffle:
The thing is, there isnt a good way to go about normalizing the numbers
because you never know how the printer is going to be setup and you dont
inherently know which way the motors will spin... At least in software...

IamTheVector:
It's not a big issue. I can make if statement based on motor direction and endstop direction or both. The fact is it seems that is not necessary. I Just tried with upper rightmost endstop by reversing motors. And it works.

IamTheVector:
I tried every homing configuration with every probe position. This is working. Tomorrow i will make a small tutorial.

boelle:

jbrazio:

PR #115, opened 2015-08-30
warn if controller is incompatible with servo library

chris-bo:
Servos need a controller with at least two 16 bit timers to work.


Wackerbarth:
There is no reason to attempt to name the particular mcu's that have the necessary timer.
The compiler provides a symbol for each timer. So, if you want this extra "hint", just check for the existence of the necessary timer symbol. That way, if some new device comes along, your check will still be correct.

chris-bo:
I copied the available options in Marlins servo.h. 

If you have any other mcu or some new device, the servo library uses timer1 as default.
However in Marlin timer1 is occupied by stepper functions so this line is commented out.
As a result the everything compiles fine, but the servos simply won't work.

If we want to keep this more generic, servo.h needs to be updated.
Maybe we could use timer3 as default and check for timer3?

chris-bo:
But can we be sure, that timer3 is always 16 bit?

boelle:

jbrazio:

PR #116, opened 2015-08-31
Implement PID_ADD_EXTRUSION_RATE

thinkyhead:
Implement this feature, as requested in https://github.com/MarlinFirmware/Marlin/issues/2604

Based on the implementation by svanteg at http://forums.reprap.org/read.php?147,364712


thinkyhead:
@Wackerbarth Placing all these PRs under a single one (#166) is not an approach we should get into the habit of following. Why not just merge each of these PRs individually? And "Candidate" shouldn't really be a PR, it should probably be done as a separate branch so that PRs in this branch can still be worked on without affecting the state of the release candidate. In point of fact, if we are going to seriously put out 1.1.0-RC1 it should be done as a tag in the Release branch of MarlinFirmware/Marlin rather than here in MarlinDev, so that it can be announced to the general public (via Twitter, Google+, forums.reprap,org, etc.) as a release candidate.

Wackerbarth:
I don't disagree and when it IS publicly released, it will have a tag, etc.  The problem is that that should have happened over a month ago. But, rather than help in doing so, you continue to work on various issues and merge them in as you see fit. Last week, I ask you to bring things up-to-date and you failed to respond directly. Since the deadline is upon us, I swept up all of the stuff that you seemed to have going and put it in a PR which I called Candidate. Basically, it is the only mechanism that I have available to draw a line which includes any of the items that you had in progress.

If you think that we need to "correct" the PR by backing out some of the changes, I am happy to do so. However, that will imply that those items will NOT be fixed in RC1.

thinkyhead:
Sorry, I have this thing called "a life" that keeps getting in the way of my Marlin duties. In fact, I have to move house this week, so expect even more absences.

Issue #122, opened 2015-09-02
Auto Bed Leveling does not work with tool offsets


jminardi:
The tool offsets are not taken into account with auto bed leveling. The following patch seemed to fix the issue for me:

```patch
index 72334d2..8dcfa2a 100644
--- a/Marlin/Marlin_main.cpp
+++ b/Marlin/Marlin_main.cpp
@@ -5069,9 +5069,20 @@ inline void gcode_T() {
             delayed_move_time = 0;
           }
         #else // !DUAL_X_CARRIAGE
-          // Offset extruder (only by XYZ)
-          for (int i=X_AXIS; i<=Z_AXIS; i++)
-            current_position[i] += extruder_offset[i][tmp_extruder] - extruder_offset[i][active_extruder];
+          // Offset extruder (only by XYZ) and make sure the bed level rotation matrix is applied.
+          vector_3 tmp_offset_vec = vector_3(extruder_offset[X_AXIS][tmp_extruder],
+                                             extruder_offset[Y_AXIS][tmp_extruder],
+                                             extruder_offset[Z_AXIS][tmp_extruder]);
+          vector_3 act_offset_vec = vector_3(extruder_offset[X_AXIS][active_extruder],
+                                             extruder_offset[Y_AXIS][active_extruder],
+                                             extruder_offset[Z_AXIS][active_extruder]);
+          vector_3 offset_vec = tmp_offset_vec - act_offset_vec;
+          offset_vec.apply_rotation(plan_bed_level_matrix);
+          current_position[X_AXIS] += offset_vec.x;
+          current_position[Y_AXIS] += offset_vec.y;
+          current_position[Z_AXIS] += offset_vec.z;
+
+
           // Set the new active extruder and position
           active_extruder = tmp_extruder;
         #endif // !DUAL_X_CARRIAGE
```

All it does it apply the same rotation matrix to the tool offset

thinkyhead:
Thanks! I will put this into a PR shortly…

PR #124, opened 2015-09-02
Apply bed leveling matrix when switching extruders

thinkyhead:
As noted in #122 by @jminardi, the auto bed leveling matrix is not taken into account when switching extruders. This patch applies the corrected difference between extruders.


jminardi:
:+1:

Roxy-3DPrintBoard:
I'm trying to get my Dual Extruder machine going...   So I'm going to be looking at this issue first hand!   I am having so many troubles getting Slicer to put out the right GCodes!

Roxy-3DPrintBoard:
Has anybody tried these changes on a Delta ?   I'm starting to look at this issue.   And I'm not sure that these changes are going to handle what I'm seeing.

Roxy-3DPrintBoard:
@AnHardt & @Thinkyhead  :   I'm fighting some horrible problems on my Delta.   I really need to get the bed leveled better and I may need to reprint the effector with some changes so I can get the nozzles at exactly the right height.   But I wanted to get you two thinking about this bug fix.

I haven't given this enough thought to know this for sure.  But I think this bug fix is only reasonable for Cartesian printers.   What I'm thinking is for Delta printers what has to happen is the  bed_level[][] matrix needs to be queried for the currently extrapolated offset based on which extruder is active.   Just because you switch extruders doesn't mean the effector should jump up or down.  And that is happening right now.

Marlin_main.cpp
```diff
@@ -5625,9 +5625,24 @@ inline void gcode_T(uint8_t tmp_extruder) {
...
+          #if ENABLED(AUTO_BED_LEVELING_FEATURE)
+            // Offset extruder, make sure to apply the bed level rotation matrix
+            vector_3 tmp_offset_vec = vector_3(extruder_offset[X_AXIS][tmp_extruder],
+                                               extruder_offset[Y_AXIS][tmp_extruder],
+                                               extruder_offset[Z_AXIS][tmp_extruder]),
+                     act_offset_vec = vector_3(extruder_offset[X_AXIS][active_extruder],
+                                               extruder_offset[Y_AXIS][active_extruder],
+                                               extruder_offset[Z_AXIS][active_extruder]),
+                     offset_vec = tmp_offset_vec - act_offset_vec;
+            offset_vec.apply_rotation(plan_bed_level_matrix);
```
jminardi: I just realized I pasted the wrong code when demonstrating this fix.

The `plan_bed_level_matrix` needs to be transposed here before applying the rotation.

See here: https://github.com/Voxel8/Marlin/blob/master/Marlin/Marlin_main.cpp#L6146

@Roxy-3DPrintBoard, maybe this is the issue you are seeing?

Roxy-3DPrintBoard:
@jminardi 
> The plan_bed_level_matrix needs to be transposed here before applying the rotation.
> See here: https://github.com/Voxel8/Marlin/blob/master/Marlin/Marlin_main.cpp#L6146
> @Roxy-3DPrintBoard, maybe this is the issue you are seeing?

No.  I don't think this is the issue I'm seeing.   The Delta printers do not do a 'Tilted Bed' correction.   Instead, it is more like the mesh leveling where it looks at the table of values and interpolates what to do to flatten the movement.

I want to caution you that I have not looked at this enough to know for sure, but I think this code should not be part of what Delta's do.  I think we need to correct based on where the effector is and not where the nozzle is.  But like I say...  I need to dig in more on this so I'm more than willing to be corrected if something I'm saying doesn't make sense!   :)

thinkyhead:
Ok, to make this fix properly, according to @jminardi just one change needs to be done. The line which reads:
```cpp
offset_vec.apply_rotation(plan_bed_level_matrix);
```
…should be changed to…
```cpp
offset_vec.apply_rotation(plan_bed_level_matrix.transpose(plan_bed_level_matrix));
```
I will make a new set of PRs for `dev` and `RCBugFix` that make this change as specified.

Issue #114, opened 2015-08-30
LCD #Move Axis way to slow for delta z-axis


o-lampe:
When I use the smart LCDs "move axis" menue, for z-moves it is way slower than  for x and y axis.
For a delta this is a PITA. Since deltas use different configuration.h and configuration_adv.h, it would be beneficial to define a speed setting there.

Cheers
o_lampe

thinkyhead:
Good idea. I'm putting this on my to-do list. We've had some other requests related to the speed of delta Z movement, homing, and probing.

o-lampe:
I´m just chewing through the newest dev files and found the right line in config_adv.h
 #if ENABLED(ULTIPANEL)
  #define MANUAL_FEEDRATE {50*60, 50*60, 4*60, 60} // Feedrates for manual moves along X, Y, Z, E from panel

This was copied from the "example configuration/ generic delta subfolder. Should be set to 50*60 for all three axis´.

thinkyhead:
Good catch, @o-lampe thanks!

PR #128, opened 2015-09-03
Remove extra Allenkey examples from Delta configs

thinkyhead:
As noted by @o-lampe at https://github.com/MarlinFirmware/MarlinDev/issues/114#issuecomment-136757614 the `MANUAL_FEEDRATE` for Delta Z should be the same as XY.


thinkyhead:
> Wouldn't it be better…?

Indeed, and done!

taalabash:
Thank you for this really fast fix!!  :)

clefranc:
@thinkyhead I can't understand why there is 3 copy of the Z_PROBE_ALLEN_KEY deploy and stow in the delta configuration files. One generic, one for Kossel mini and one for pro...

thinkyhead:
@clefranc Originally there was only one delta configuration. As these were split up by the contributors, some of them altered the existing options while others simply un-commented the one appropriate to the specific robot. At present you can think of all the ones that are commented-out as alternative examples. But, I'm all for having only a single set of options and removing the commented-out ones, since that's the point of the examples — to give the most likely default settings for a specific design.

Wackerbarth:
The remaining changes here now have nothing to do with feed rate.

All that is left here is the removal of some Allen Key movement specifications.

These should NOT be removed until we address the greater issue of how we handle ALL similar configuration situations.

This should be done as a part of the reorganization for Marlin 1.2

In particular, although I agree that the individual examples should not have all of the alternatives, we do need to have a reference document which has the information appropriate to assist a user in configuraing his own.

thinkyhead:
Cleaning up lingering branches… The name of this PR has been changed, and it's still the right idea.

Someone else will need to create a new PR for this, as my fork will not be accepting any commits that were made after September 14th.

Roxy-3DPrintBoard:
> thinkyhead commented 26 days ago
>@clefranc Originally there was only one delta configuration. As these were split up by the contributors, some of them altered the existing options while others simply un-commented the one appropriate to the specific robot. At present you can think of all the ones that are commented-out as alternative examples.

The big problem with the AllenKey stuff is it only handles specific implementations right now.  One big weakness is there is no concept of the AllenKey having to be deployed (and retracted) manually.  I'm bringing up the Release Candidate on my new G2s Delta kit and having to make changes to handle that issue.

clefranc:
@Roxy-3DPrintBoard Resurrecting this *old* thread:
https://github.com/MarlinFirmware/MarlinDev/issues/61#issuecomment-137518095

Basically, you have the same manual dock/stow probe than me. Solved this issue with added start G-code in my slicer.

Wackerbarth:
Superseded -- The removal of the inactive key motion sequences is a natural byproduct of the new configuration approach.

PR #132, opened 2015-09-04
Fix Z_PROBE_ALLEN_KEY section

clefranc:
- Deleted Z_PROBE_ALLEN_KEY section from example_configurations\delta\biv25 (not used)
+ Added Z_PROBE_ALLEN_KEY_DEPLOY_3 to example_configurations\delta\generic and \kossel_mini (undefined)
- Deleted Kossel Mini and Kossel Pro Z_PROBE_ALLEN_KEY from example_configurations\delta\generic
- Deleted Kossel Pro Z_PROBE_ALLEN_KEY from example_configurations\delta\kossel_mini
- Deleted Kossel Mini Z_PROBE_ALLEN_KEY from example_configurations\delta\kossel_pro


Wackerbarth:
This change is in error. The DEPLOY_3 parameters are INTENTIONALLY missing because that configuration is using a 2 step deployment rather than a 3 step deployment.

Also, although I agree that the unused code SHOULD be removed, the present policy is that every configuration maintains the alternatives (commented out) so that, in theory, you could start from any configuration and switch to another by changing choice of lines which are enabled.

PR #138, opened 2015-09-04
Fix for stutter due to timer discrepancy

thinkyhead:
As noted in MarlinFirmware/Marlin#2525 a 16-bit timer wraps around a lot, so you need to be extra careful with time comparisons.


oysteinkrog:
I tested this on my machine and I can confirm it works. This PR needs to be cleaned up though, I only tested 76e9a2c (cherry-picked into my branch).
For me this is a _huge_ improvement, as it means I can finally use high speeds.

Wurstnase:
I've got similar issues with 32bit-timer on the due. But instead of ~32ms I wait ~120s.
The problem I think is, that marlin calculates it's timer in the ISR but in the end.

Wackerbarth:
> On Sep 10, 2015, at 3:10 AM, Wurstnase <notifications@github.com> wrote:
> 
> I've got similar issues with 32bit-timer on the due. But instead of ~32ms I wait ~120s.
> The problem I think is, that marlin calculates it's timer in the ISR but in the end.
> 
I don’t think that the problem is where, or how, the timer is calculated, but how much time it takes to process the interrupt.
Basically, by the time the ISR finishes, the time interval until the next interrupt is now negative.

This proposed change replaces that negative interval with a very short time (right away). The net effect is that we get an unregulated slow down in the interrupt frequency.

There are really only three possibilities to solve the problem in a controlled way.

1) Perform fewer calculations per step in the ISR,
2) Use a faster MPU, or
3) Lower the maximum step speed.

The double/quad stepping strategy is already used to effectively lower the calculation per step.
And some users have indicated that they have been successful by lowering maximum stepper frequency.

I suggest that we incorporate a combination of this time skewing hack with some alarm indication which gets triggered when it comes into play.

Wurstnase:
I think you are right. All calculations in an interrupt should have time to be calculated. But this is also very dependend on some setups. e.g. endstop not only for homing...

Wackerbarth:
Perhaps we need to rethink the handling of end stops.  They serve two purposes. During "homing" and "probing", they serve as alignment sensors. In this case, we need accurate sensing of the stepper position at the time that the end stop triggers.
Their other use is as a safety. A properly running bot should never trigger the end stop unexpectedly.

I will suggest that, for this purpose, the accuracy of the trigger point is not an issue.

In the former case, we can, and should, approach the trigger point at a slow speed. By doing so, we can assure that the end stop is sampled more frequently than the stepping rate.

In the latter case, when the alarm condition is triggered, we really need to abort the printing operation. If we are going to do so, it should not matter if a small number of additional steps have occurred since the end stop trigger.

So, perhaps we could remove the end stop testing from the primary stepping loop and allow it to operate at a lower frequency.

PR #144, opened 2015-09-06
Fix up probe pin define and test

thinkyhead:


thinkyhead:
After rebase it appears this is redundant.

PR #151, opened 2015-09-07
Introduction of Sımple Outputs For Spindle Control Gcodes

msutas:
This change introduces M3, M4 and M5 gcodes to Marlin which may be used to control a spindle or laser unit connected to the machine


clefranc:
@msutas Don't want to bother you with this, but I think your PR is a little incomplete:

The new `#define` in `Configuration.h` should be added to the other configuration files as well (in `example_configurations` and `configurator/config` folders).

Also, you should add warning(s) compile in `SanityCheck.h` to `SPINDLE_CONTROL` for proper `SPINDLE_DIR_PIN` and `SPINDLE_SPEED_PIN` usage.

Marlin/Marlin_main.cpp
```diff
@@ -2916,6 +2926,30 @@ inline void gcode_G92() {
...
+   */
+  inline void gcode_M4() { spindleSpeed = code_seen('S') ? constrain(code_value_short(), 0, 255) : 255;
+                           digitalWrite(SPINDLE_DIR_PIN,HIGH);
+                           analogWrite(SPINDLE_SPEED_PIN, spindleSpeed); }
+  
+  /**
+   * M5: Stop Spindle
+   */
+  inline void gcode_M5() { digitalWrite(SPINDLE_DIR_PIN,LOW);
+                           analogWrite(SPINDLE_SPEED_PIN, 0); }
```
CONSULitAS: shouldn't you reset the DIR_PIN also?

`digitalWrite(SPINDLE_DIR_PIN,LOW);`

Marlin/Marlin_main.cpp
```diff
@@ -525,6 +528,12 @@ void setup_filrunoutpin() {
   #endif
 }
 
+void setup_spindlepin() {
+  #if HAS_SPINDLE
+    pinmode(SPINDLE_DIR_PIN,OUTPUT);
+  #endif
```
CONSULitAS: shouldn't you init the SPEED_PIN also?

`pinmode(SPINDLE_SPEED_PIN,OUTPUT);`

msutas:
@CONSULitAS it may not be necessary to initialize these pins because each code (M3, M4, M5) sets the pins. But to be on the safe side I added it to the setup.

msutas:
@clefranc I am updating the files as your suggestion but I don't find it meaningful to update all configuration file examples which are mostly for commercial products. I believe in an open source project like this the owner of the commercial product should keep track of the changes and update the example configuration related to his machine when needed. Current process is a burden for the contributor. What will happen if 50 more example configurations are created in the future?

fmalpartida:
@msutas +1. 
There are many ways to support their customers by simply forking and maintaining their code base. I would really hate to become the help service for such bots.

Wackerbarth:
The difficultly is the way that the present configuration is maintained. I have been advocating an alternate procedure for some time.

Unlike those who dislike the idea of our providing a source of customer support, I remind you that there are many vendors who fail to provide continuing support, often because they are no longer in business. Their, now orphaned, customers come to the Open Source community for support. Therefore, I contend that it IS in our interest to be able to retrieve the configuration information for ANY machine based on our firmware.

The thing that we need to do differently is make that process as convenient for both the vendor and ourselves.  An alternate approach to maintaining the Configuration files can go a long way toward providing the needed mechanisms.

boelle:

msutas:
@boelle OK I am just closing this pull request and will create a new pull request on the main repro in the following days.

boelle:
thanks....

PR #143, opened 2015-09-06
Fix compile error in lcd_move_e calls

thinkyhead:
Fix a compile error introduced in #98 due to missing semicolons.

PR #98, opened 2015-08-20
Extrude From Multiple Extruders On an LCD Controller

gralco:
Title says it all.

PR #154, opened 2015-09-09
Extrude From Multiple Extruders Changes

gralco:
Fixes for https://github.com/MarlinFirmware/MarlinDev/pull/98 per https://github.com/MarlinFirmware/MarlinDev/pull/143.

### DIFF
--- a/Marlin/language_en.h
+++ b/Marlin/language_en.h
@@ -117,22 +117,27 @@
 #ifndef MSG_MOVE_Z
 #define MSG_MOVE_Z                          "Move Z"
 #endif
-#ifndef MSG_MOVE_E
-#define MSG_MOVE_E                          "Extruder"
+#if EXTRUDERS == 1
+  #ifndef MSG_MOVE_E
+  #define MSG_MOVE_E                        "Extruder"
+  #endif
 #endif
 #if EXTRUDERS > 1
   #ifndef MSG_MOVE_E1
   #define MSG_MOVE_E1                       "Extruder1"
   #endif
+  #ifndef MSG_MOVE_E2
+  #define MSG_MOVE_E2                       "Extruder2"
+  #endif
   #if EXTRUDERS > 2
-    #ifndef MSG_MOVE_E2
-    #define MSG_MOVE_E2                     "Extruder2"
+    #ifndef MSG_MOVE_E3
+    #define MSG_MOVE_E3                     "Extruder3"
     #endif
     #if EXTRUDERS > 3
       #ifndef MSG_MOVE_E3
-      #define MSG_MOVE_E3                   "Extruder3"
+      #define MSG_MOVE_E3                   "Extruder4"
       #endif
-    #endif
+    #endif // EXTRUDERS > 3
   #endif // EXTRUDERS > 2
 #endif // EXTRUDERS > 1
 #ifndef MSG_MOVE_01MM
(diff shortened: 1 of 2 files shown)


gralco:
@clefranc @thinkyhead

KiteLab:
All that conditions in the language file are useless. Ether the defined strings are used, then they are compiled in, or not. If they are defined, but not used, nothing bad will happen.

The same is valid for the DELTA_CALIBRATION_MENU.

shampine1:
Can we merge this?  Is there anything I can do (with my limited programming skills) to help?  I can't get my full graphic lcd to work at all!  And, of course, I can't compile it with 2 extruders...

Wackerbarth:
@KiteLab is correct. There is no reason to have any conditional definition of the Messages themselves.
Further, the code can be simplified by defining the message for the first extruder as one of two fixed messages in the Conditionals. Then the code just adds one menu item for each of the extruders.

shampine1:
@gralco , I get a lot of errors when I try to compile your branch with 1 or 2 extruders, with and without "#define REPRAP_DISCOUNT_FULL_GRAPHIC_SMART_CONTROLLER".  I see 2 unterminated if statements that I experimentally fixed, and a lot of the errors below.  Also, I see an "expected unqualified-id before '<<' token that lights up "<<<<<<< HEAD".  I don't know enough programming to know what the string of less than symbols is supposed to do.

The unterminated #if statements are at line 2 and line 369.  I'm pretty sure the latter one just needs an endif 3 lines down, but I'm not sure what the one at line 2 needs.

ultralcd.cpp:568: warning: 'void lcd_preheat_pla_menu()' defined but not used
ultralcd.cpp:592: warning: 'void lcd_preheat_abs_menu()' defined but not used
ultralcd.cpp:737: warning: 'void lcd_move_menu_axis()' declared 'static' but never defined
ultralcd.cpp:752: warning: 'void lcd_move_x()' defined but not used
ultralcd.cpp:753: warning: 'void lcd_move_y()' defined but not used
ultralcd.cpp:754: warning: 'void lcd_move_z()' defined but not used
ultralcd.cpp:785: warning: 'void lcd_move_e0()' defined but not used

KiteLab:
There is no "<<<<<<< HEAD" in the original code of Marlin. It's a result of a failed merge.
Please re-download.

gralco:
@KiteLab 
Oh, that's really nice that the compiler handles it. I made the changes.
@shampine1 
I used `grep` to search for conflict markers and I found none in my code. If you find any then please point me to it.
@Wackerbarth 
Could you please show me an example of what you mean?

shampine1:
Gralco,  That was quick!
I'm getting "missing binary operator before token "("" errors in ultralcd now; it high lights line 11, and complains about 11, 16, and 448.  Line 11 seems to be the same as the one in the current development branch, so I don't know why it doesn't compile.

thinkyhead:
There are other issues with this PR, such as a missing definition for `MSG_MOVE_E4`. And actually, we've been sticking to the convention of using E0 as the name of the first extruder, because that's what is silk-screened onto most boards. See #161 for a PR that supplants this one, with slightly smaller code.

----

In the process of testing I also noticed that pin overrides in `Configuration_adv.h` are conflicting with existing pin definitions in `pins_AZTEEG_X3_PRO.h` (and certainly other files). Probably these pin overrides need to be commented out. The convention of overriding pins in `Configuration_adv.h` is problematic, because `#define` is processed as files are loaded, not necessarily in the position where `#include` falls. This problem has arisen because we now define the default pins for each board, so the existence of an optional pin can no longer be assumed to mean "enable this feature."

PR #155, opened 2015-09-10
RAMPS_13_EEB doesn't exist

shampine1:
The EEB configuration (Extruder, Extruder, Bed) isn't set up right; the third output is being set to a fan by default.  I added two lines to fix this.


KiteLab:
Rubbish!
The fan is associated with pin 4 (what is the right most servo pin - a good choice for the RRD Fan Extender when placed the better way - shifted two pins to the right to leave the first two servo pins open).
In line 113 the bed gets its normal pin (8).

shampine1:
@Kitelab,
Thanks for your review!  After a lot more head-scratching, I understand what you mean.  You are right, the EEB case is already handled right!

Sorry!

PR #159, opened 2015-09-11
Create language_cz.h

clexpert:
Czech language translation.

PR #161, opened 2015-09-11
Fix extruder-related LCD menu items

thinkyhead:
As noted by @gralco in #143, we also need this to resolve the compile error.

- This is similar to #154 but reduces code size slightly.
- Cleans up menu items related to single vs. multiple extruders.
- Changes human-readable messages so they are numbered 1-4 instead of 0-3.
- Update language files for the new naming scheme.
- Adds Czech language from #159 by @clexpert with language changes.

### DIFF
--- a/Marlin/Configuration_adv.h
+++ b/Marlin/Configuration_adv.h
@@ -41,7 +41,8 @@
   // if Kc is chosen well, the additional required power due to increased melting should be compensated.
   #define PID_ADD_EXTRUSION_RATE
   #if ENABLED(PID_ADD_EXTRUSION_RATE)
-    #define  DEFAULT_Kc (1) //heating power=Kc*(e_speed)
+    #define DEFAULT_Kc (100) //heating power=Kc*(e_speed)
+    #define LPQ_MAX_LEN 50
   #endif
 #endif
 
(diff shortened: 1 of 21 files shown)


thinkyhead:
> I find these inconsistencies a little annoying.

Yes, they are. I support using 1-4 since that is the most human-readable convention. But we'll need to apply it all over in the human-readable output, and then the question is whether to leave the names of related settings numbered 0-3.

thinkyhead:
Ok, so https://github.com/thinkyhead/MarlinDev/commit/46ec686556f00de8e1d50a35367910820e41b5be applies consistent naming through all the menu items. With a single extruder / heater / nozzle the menu shows no number on the end of items. With more than 1 extruder it appends a number from 1 to 4. This applies only to the LCD menu so far. Serial output messages for things like M303, GCode commands that take an extruder index, etc., continue to use 0 as the first index.

clefranc:
@thinkyhead Thank you!

thinkyhead:
Alright… after a lot of patching up, rebasing, combining, squashing… this should be pretty complete. I just looked for `EXTRUDERS > 2` throughout `ultralcd.cpp` and fixed up each block that uses it.

thinkyhead:
Where is the PR which supersedes this one?

PR #163, opened 2015-09-12
Disable DEBUG_LEVELING_FEATURE by default

thinkyhead:
This feature is a huge memory hog. Will it be removed for the release or left in? If it will be left in, it should be disabled by default.


Roxy-3DPrintBoard:
Agreed to Don't Enable Debug Leveling Feature by default.   But most real
printers have an Auto Bed Leveling mechanism.   It makes sense to have that
enabled by default.   Right?

Wackerbarth:
It was never intended that DEBUG_LEVELING_FEATURE be enabled by default.

Note that your PR both enables and then disables it.

PR #164, opened 2015-09-12
Conditionalize functions that rely on SDSUPPORT

thinkyhead:
Addresses compiler warnings that have emerged now that `SDSUPPORT` must be manually enabled.


thinkyhead:
The error from Travis is only because originally this branch was pushed to `upstream` instead of `origin`. Otherwise it's fine. I'll see if I can patch up my remote to use the correct origin.

thinkyhead:
Where is the PR which supersedes this one?

Issue #135, opened 2015-09-04
RRD Smart LCD 2004 "no SD-Card" error


o-lampe:
When I power On my Ramps with above mentioned LCD and the SD-Card is already inserted, I see " no SD-Card" message in main display. I have to remove and plug it back in to have access to SD-files. 
Not a biggy, but with autostart option enabled, it wouldn´t work.

Wackerbarth:
There was an issue about this which was raised earlier and I thought that it had been fixed. Are you using the latest, or at least a very recent, version of the dev firmware?

o-lampe:
@Wackerbarth  I´ve downloaded the last version two days ago, maybe I haven´t set all the right options?
Usually I disable #define SDsupport
and enable RRD_Smart Controller only. Worked fine in the past.

thinkyhead:
Going forward you will have to enable `SDSUPPORT` explicitly in `Configuration.h`. That shouldn't have any effect on how it behaves, however. If you have `SD_DETECT_PIN` defined (and the correct `SD_DETECT_INVERTED` setting) then Marlin should automatically detect the presence of an already-inserted SD card. If it's not, maybe there's some way to twiddle the detection pin to cause it to refresh its state. Just spitballing…

alexandrezia:
I have the same issue here, On boot it doesn't detect an already inserted SD card
on latest commit  b23d7659 
using board:  #define REPRAP_DISCOUNT_SMART_CONTROLLER
Played with all combinations of SDSUPPORT and SD_DETECT_INVERTED.

o-lampe:
@thinkyhead  enabling SDSUPPORT didn´t help. Don´t know where the other variables you mentioned are located. In the past, I never had to change anything outside the configuration.h to make a printer work.

thinkyhead:
@alexandrezia Try this as a test. Replace the line which reads `lcd_oldcardstatus = IS_SD_INSERTED;` with `lcd_oldcardstatus = false;` and see if it causes the initialization routine to do the right thing. (Namely, to see a change in status.)

o-lampe:
`<addr>`
#if ENABLED(SDSUPPORT) && PIN_EXISTS(SD_DETECT)
    pinMode(SD_DETECT_PIN, INPUT);
    WRITE(SD_DETECT_PIN, HIGH);
    lcd_oldcardstatus = false;
  #endif
`<addr>`

[...]

PR #165, opened 2015-09-12
Init lcd_sd_status as false to force lcd_update test

thinkyhead:
Replaces #145 which is way out of date

As noted in #135 the initial test for the presence of an SD card may be preventing `lcd_update` from calling `card.initsd`.

An alternative to this could call `card.initsd` in `lcd_init` instead.


thinkyhead:
@Wackerbarth I wouldn't merge this code in yet until the issue mentioned in https://github.com/MarlinFirmware/MarlinDev/issues/135#issuecomment-138761448 is resolved. Specifically:

> The only thing I've noted is that as mentioned by @o-lampe it overwrites the last line since boot and I can't check the "user defined text" that usually sits on that last line.

PR #170, opened 2015-09-12
Fix home/quickhome for corexy/xz

darkjavi:
Coming from #57 , https://github.com/MarlinFirmware/Marlin/pull/2378,  https://github.com/MarlinFirmware/Marlin/pull/2255.

Resume: There is a bug in the coreXY/XZ homing implementation, that is only exposed when enabling QUICK_HOME. It fetches raw steps (a/b) from the sttepers instead of axis movement (x/y). 

This is another approach to the issue,same essence but this time using just bools on ISR time.

@thinkyhead I think you wanted included the float helpers and the rawX/Y/Z movement. I included the helper functions but removed the floats with raw dx/dy/dz from the block as its now done with a bool I don't have any inmediate use for them.

While doing the fix was thinking if it might have sense to have the is_axis_moving[] array enabled in all configurations, but then again doesn't have an immediate use beyond fixing the core issue.

### DIFF
--- a/Marlin/macros.h
+++ b/Marlin/macros.h
@@ -1,6 +1,8 @@
 #ifndef MACROS_H
 #define MACROS_H
 
+#include <math.h>
+
 // Macros for bit masks
 #define BIT(b) (1<<(b))
 #define TEST(n,b) (((n)&BIT(b))!=0)
@@ -23,5 +25,10 @@
 #define DISABLED(b) (!_CAT(SWITCH_ENABLED_, b))
 
 #define COUNT(a) (sizeof(a)/sizeof(*a))
+
+// float helpers
+#define EPSILON ((float)0.0000001)
+inline bool nearly_zero(float x) { return fabs(x) < EPSILON; }
+inline bool nearly_equal(float x, float y) { return nearly_zero(x - y); }
 
 #endif //__MACROS_H
--- a/Marlin/planner.cpp
+++ b/Marlin/planner.cpp
@@ -544,11 +544,19 @@ float junction_deviation = 0.1;
     block->steps[A_AXIS] = labs(dx + dy);
     block->steps[B_AXIS] = labs(dx - dy);
     block->steps[Z_AXIS] = labs(dz);
+    block->is_axis_moving[X_AXIS] = !nearly_zero(dx);
+    block->is_axis_moving[Y_AXIS] = !nearly_zero(dy);
+    block->is_axis_moving[Z_AXIS] = !nearly_zero(dz);
+    block->is_axis_moving[E_AXIS] = !nearly_zero(de);    
   #elif ENABLED(COREXZ)
     // corexz planning
     block->steps[A_AXIS] = labs(dx + dz);
     block->steps[Y_AXIS] = labs(dy);
     block->steps[C_AXIS] = labs(dx - dz);
+    block->is_axis_moving[X_AXIS] = !nearly_zero(dx);
+    block->is_axis_moving[Y_AXIS] = !nearly_zero(dy);
+    block->is_axis_moving[Z_AXIS] = !nearly_zero(dz);
+    block->is_axis_moving[E_AXIS] = !nearly_zero(de);
   #else
     // default non-h-bot planning
     block->steps[X_AXIS] = labs(dx);
@@ -1039,3 +1047,4 @@ void reset_acceleration_rates() {
   for (int i = 0; i < NUM_AXIS; i++)
     axis_steps_per_sqr_second[i] = max_acceleration_units_per_sq_second[i] * axis_steps_per_unit[i];
 }
+
--- a/Marlin/stepper.cpp
+++ b/Marlin/stepper.cpp
@@ -309,13 +309,18 @@ inline void update_endstops() {
   // TEST_ENDSTOP: test the old and the current status of an endstop
   #define TEST_ENDSTOP(ENDSTOP) (TEST(current_endstop_bits, ENDSTOP) && TEST(old_endstop_bits, ENDSTOP))
 
-  #define UPDATE_ENDSTOP(AXIS,MINMAX) \
+  #define _UPDATE_ENDSTOP(AXIS,MINMAX,FIELD) \
     SET_ENDSTOP_BIT(AXIS, MINMAX); \
-    if (TEST_ENDSTOP(_ENDSTOP(AXIS, MINMAX))  && (current_block->steps[_AXIS(AXIS)] > 0)) { \
+    if (TEST_ENDSTOP(_ENDSTOP(AXIS, MINMAX)) && current_block->FIELD[_AXIS(AXIS)] > 0) { \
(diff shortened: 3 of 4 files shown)


Marlin/stepper.cpp
```diff
@@ -309,13 +309,22 @@ inline void update_endstops() {
...
+        step_events_completed = current_block->step_event_count;  }
+  #else
+    #define UPDATE_ENDSTOP(AXIS,MINMAX) \
+      SET_ENDSTOP_BIT(AXIS, MINMAX); \
+      if (TEST_ENDSTOP(_ENDSTOP(AXIS, MINMAX))  && (current_block->steps[_AXIS(AXIS)] > 0)) { \
+        endstops_trigsteps[_AXIS(AXIS)] = count_position[_AXIS(AXIS)]; \
+        _ENDSTOP_HIT(AXIS); \
+        step_events_completed = current_block->step_event_count;  }
+  #endif
+
```
thinkyhead: This can actually be reduced to eliminate all those redundant lines. Here's the replacement code:
```cpp
  #define _UPDATE_ENDSTOP(AXIS,MINMAX,FIELD) \
    SET_ENDSTOP_BIT(AXIS, MINMAX); \
    if (TEST_ENDSTOP(_ENDSTOP(AXIS, MINMAX)) && current_block->FIELD[_AXIS(AXIS)] > 0) { \
      endstops_trigsteps[_AXIS(AXIS)] = count_position[_AXIS(AXIS)]; \
      _ENDSTOP_HIT(AXIS); \
      step_events_completed = current_block->step_event_count; \
    }
  #if ENABLED(COREXY) || ENABLED(COREXZ)
    #define UPDATE_ENDSTOP(AXIS,MINMAX) _UPDATE_ENDSTOP(AXIS,MINMAX,is_axis_moving)
  #else
    #define UPDATE_ENDSTOP(AXIS,MINMAX) _UPDATE_ENDSTOP(AXIS,MINMAX,steps)
  #endif
```

Marlin/planner.cpp
```diff
@@ -544,11 +544,19 @@ float junction_deviation = 0.1;
     block->steps[A_AXIS] = labs(dx + dy);
     block->steps[B_AXIS] = labs(dx - dy);
     block->steps[Z_AXIS] = labs(dz);
+    block->is_axis_moving[X_AXIS] = !nearly_zero(dx);
+    block->is_axis_moving[Y_AXIS] = !nearly_zero(dy);
+    block->is_axis_moving[Z_AXIS] = !nearly_zero(dz);
+	block->is_axis_moving[E_AXIS] = !nearly_zero(de);
```
thinkyhead: Oops, a tab slipped through. Should be 2 spaces.

thinkyhead:
Make those small tweaks and this will be merged asap!

darkjavi:
Done and tested, its working for my corexy. Can test it too on a cartesian(the dual carriage machine) but it will take some hours for his current printing job to finish.
Nice tweak of the macro, I'm learning such a lot of cool preprocessor stuff looking at marlin's code :)

thinkyhead:
@darkjavi Are you comfortable with `git` on the command line? If so, you can squash some commits pretty easily with `git rebase -i HEAD~5`. Your default text editor will open. Change `pick` on every commit but the first one to `squash` or just `s`, then save-and-close the editor. It will open again, this time to let you edit the commit log. Delete all commit messages except the first one. This time when you save, git will complete the rebase, combining all your commits into one! Then you can just use `git push -f` to upload the new commit.

If you're not comfortable with this, that's fine. I might be able to check this out and squash your commits for you. I'm just starting to get the hang of it myself…

Marlin/macros.h
```diff
@@ -24,4 +26,9 @@
 
 #define COUNT(a) (sizeof(a)/sizeof(*a))
 
+// float helpers
+#define EPSILON ((float)0.0000001)
+inline bool nearly_zero(float x) { return fabs(x) < EPSILON; }
+inline bool nearly_equal(float x, float y) { return nearly_zero(x - y); }
```
Wurstnase: This should be handeled with care. Is this everytime <0.0000001?

darkjavi:
@thinkyhead  Nice trick! I wondered myself a few times how to do that so if realize you messed up after commiting you don't need to generate more commits with 'fix for the fix'.

@Wurstnase  What do you mean? Might  be values < EPSILON?

Wurstnase:
http://floating-point-gui.de/errors/comparison/

Wackerbarth:
A much better "solution" is to reformulate the logic so that floating comparisons are not needed.

Wurstnase:
@Wackerbarth Good point.
To point to the planner:
target[_AXIS] is long
position[_AXIS] is long
dx, dy ,dz is float? Why?

Wackerbarth:
I have no good explanation for the decision to perform the calculations in the present manner.
And, given our underpowered AVR processors, I think that we should return to performing the calculations in the stepper coordinate system rather than doing floating point calculations in the object space.  But such changes will have to come during a rewrite of the kinematics.

Wurstnase:
I think we can replace ```float``` with ```long``` in that case. There are some calculations in that function where we get an automatic conversion to float because ```long/float == float```. It's worth a try.

Wurstnase:
Just start a print with my due-version on my CoreXY with ```long``` and it seems to work.

Wurstnase:
I made a PR for that issue.
https://github.com/MarlinFirmware/MarlinDev/pull/188

Wurstnase:
There are no floats anymore.

thinkyhead:
#358 should also help this. Apparently another rebase is needed. This needs to be rebased against `dev` and then re-submitted against the `dev` branch. At least, I assume so. That's the branch I keep making PRs for. I've had really good luck just doing `git rebase -i dev; git push -f` — Then you would need to close this PR and create a new one.

darkjavi:
I've been under a heavy load lately, soon to be relieved. Will take a look to the issue ass soon as can get some time.

@thinkyhead  Does #358  change the conditions to tell if an endstop has been triggered?

thinkyhead:
> Does #358 change the conditions to tell if an endstop has been triggered?

Well, it still assumes that an endstop has been triggered if the switch associated with it is depressed. However, it should now report the correct position when this occurs, whereas before this patch, when an endstop was triggered it would use the Stepper A raw position and produce a screwed up result.

boelle:

darkjavi:
Sorry guys, it's being such a busy time, if still wanna merge it I still want to get this fixed. Can give this a couple of hours today when I'm done with the shift.

I got somehow lost some months ago, when the group was discussing which would be the best way to colaborate, and don't know it ended.  @boelle  should I just fork the main repo and PR against RCbugfix?

Jnesselr:
Yep, that's perfectly fine. We're just trying to consolidate everything and get people on the same page.

jbrazio:

jbrazio:

PR #174, opened 2015-09-13
Consistent use of //#define

thinkyhead:
This makes all disabled defines use the same two-slashes-with-no-spaces convention. (The Travis script should still look for the regex `\/\/[ \t]*#define` anyways when doing search-and-replace to enable options, and not assume that there will or will not be spaces.)


Wackerbarth:
Restyling defines the format of all lines

PR #179, opened 2015-09-13
Fixed continuous homing when pressing home button on LCD controller.

fmalpartida:
Fix: increased debounce delay on home button for LCD controllers. The default debounce delay was too short and was causing several homing for each button press.


thinkyhead:
Good catch…. This has really been annoying me, and I couldn't find the cause!

Roxy-3DPrintBoard:
Can you tell me what the original denounce time was and what the 'working'
denounce time is?    I'm mostly just curious, but do any of these LCD
panels have membrane switches?   Because if they do, they are noisy as can
be.

fmalpartida:
More thank chatter is how the button reading is done. In the main loop it reads the button and if pressed it will home. If the delay is reached and the button is still pressed, it will queue a new home movement. 

Therefore, it is passing through there a few more times than expected. Perhaps it should be called rearm_delay. 

Enviado desde mi iPhone

Issue #123, opened 2015-09-02
Minor problem with "delta_calibration" menue on a 2004 LCD 


o-lampe:
When I send a "Calibrate_X" command and then navigate back to the main screen, the X-position shown is wrong. ( X-21 instead of X-121) The first digit got chopped off.

Newest Dev-master download.
Cheers
-Olaf

o-lampe:
Same for calibrate_Z. Display reads Y140 instead of Y-140.
This time the - sign is chopped off.

thinkyhead:
I see that the delta calibration menu is meant to be customized before flashing. The default code is pasted below. What does your delta calibration code look like?
```cpp
static void lcd_delta_calibrate_menu() {
  START_MENU();
  MENU_ITEM(back, MSG_MAIN, lcd_main_menu);
  MENU_ITEM(gcode, MSG_AUTO_HOME, PSTR("G28"));
  MENU_ITEM(gcode, MSG_DELTA_CALIBRATE_X, PSTR("G0 F8000 X-77.94 Y-45 Z0"));
  MENU_ITEM(gcode, MSG_DELTA_CALIBRATE_Y, PSTR("G0 F8000 X77.94 Y-45 Z0"));
  MENU_ITEM(gcode, MSG_DELTA_CALIBRATE_Z, PSTR("G0 F8000 X0 Y90 Z0"));
  MENU_ITEM(gcode, MSG_DELTA_CALIBRATE_CENTER, PSTR("G0 F8000 X0 Y0 Z0"));
  END_MENU();
}
```

o-lampe:
@thinkyhead 
Mine looks the same, except, I adapted it to my bigger printable radius of 140mm:
#if ENABLED(DELTA_CALIBRATION_MENU)
  static void lcd_delta_calibrate_menu() {
    START_MENU();
    MENU_ITEM(back, MSG_MAIN, lcd_main_menu);
    MENU_ITEM(gcode, MSG_AUTO_HOME, PSTR("G28"));
    MENU_ITEM(gcode, MSG_DELTA_CALIBRATE_X, PSTR("G0 F5000 X-121.24 Y-70.0 Z0"));
    MENU_ITEM(gcode, MSG_DELTA_CALIBRATE_Y, PSTR("G0 F5000 X121.24 Y-70.0 Z0"));
[...]

PR #181, opened 2015-09-14
Change XY formatting on LCD

thinkyhead:
According to #123 negative values for XY at or below -100 are displaying incorrectly, dropping the first digit. Deltas can easily have XY values in this range. This PR adds a function to display floats/ints formatted like `_123`, `-123`, `_-12`, or `__-1` as appropriate and applies it to the XY coordinates on Hitachi displays. It also moves the Z value to the right to be consistent with the XY formatting.


AnHardt:
A minor enhancement for the documentation.
```
16x4   |000/000 B000/000|
       |SD100%  Z 000.00| <
       |F100%     T--:--|
       |0123456789012345|

**Comment by thinkyhead** [comment]
Rebased. I'll probably submit this fix to the RCBugFix branch as well.

File: ultralcd.h [hunk 1]
```diff
@@ -116,21 +116,23 @@
...
-char* ftostr52(const float& x);
+char* itostr2(const uint8_t &x);
+char* itostr31(const int &xx);
+char* itostr3(const int &xx);
+char* itostr3left(const int &xx);
+char* itostr4(const int &xx);
+char* itostr4sign(const int &x);
+
+char* ftostr3(const float &x);
+char* ftostr4sign(const int &x);
```
AnHardt: int ??? --> float !!!

ultralcd.cpp
```diff
@@ -2086,6 +2087,30 @@ char* itostr4(const int& xx) {
   return conv;
 }
 
+// Convert int to rj string with _123, -123, _-12, or __-1 format
+char *itostr4sign(const int &x) {
+  long xx = abs(x);
```
AnHardt: int is save here for 4 digits.

ultralcd.cpp
```diff
@@ -1901,10 +1901,11 @@ void lcd_reset_alert_level() { lcd_status_message_level = 0; }
...
 
-// Convert float to string with +123.4 format
-char* ftostr3(const float& x) {
-  return itostr3((int)x);
-}
+// Convert float to rj string with 123 or -12 format
+char *ftostr3(const float &x) { return itostr3((int)x); }
+
+// Convert float to rj string with _123, -123, _-12, or __-1 format
+char *ftostr4sign(const int &x) { return itostr4sign((int)x); }
```
AnHardt: ftostr4sign(const int &x)
float !

AnHardt:
Replaced by #265

PR #183, opened 2015-09-14
Update temperature.cpp

12dstring:
MSG_EXTRUDER_SWITCHED_OFF was renamed to MSG_REDUNDANCY in language.h, but was not renamed here. If TEMP_SENSOR_1 is enabled as redundant then an error is thrown that MSG_EXTRUDER_SWITCHED_OFF is not defined.


Wackerbarth:
Cherry picked to dev branch

PR #188, opened 2015-09-15
dx, dy, dz and de from float -> long

Wurstnase:
they are long, so they don't need to be float.


Wurstnase:
works on my coreXY

Wurstnase:
works? works not? I can't say. I've got some issues with magically loosing steps and maybe this was the issue.

Wurstnase:
I'm a rookie... It was a mechanical issue. So this should work.

Wackerbarth:
Replaced by #297

PR #189, opened 2015-09-15
better names for delta speeds

Wurstnase:
just rename dx to dsx. its not a delta for x. its a delta for the speed of x.


Wackerbarth:
Already merged

PR #194, opened 2015-09-16
Add support for G20/G21 and inch mode

reidrankin:
Inch mode is common in generated G-code for CNC machines. This approach to integrating an inch mode avoids having to update any internal algorithms or constants by converting values as they are read in from the G-code file.


reidrankin:
This PR contains only the changes needed to support inch mode, but you can see how I'm moving towards a system that is aware of what type of value is being read more generally. For example, this PR also fixes several bugs regarding the handling of the M200 code's volumetric mode: it is supposed to change the E-axis to volumetric units, but was only doing so in movement commands executed by the planner; other commands, like M205 feedrate changes, M207/M208 retraction values, M600 filament changes, and T0-T3 feedrate changes were all still counting on linear units.

I'm working on a more comprehensive solution, replacing code_value(), code_value_short(), and code_value_long() with (inlined) functions specific to the type of value being requested, for example temperature, time, or a plain scalar value. I'll put that all in another PR derived from this branch.

reidrankin:
I've done so much work improving the code in this PR as part of my aforementioned overhaul of the code_value functions that I'm just going to close this and submit a new one for the whole kit-'n-caboodle.

PR #196, opened 2015-09-17
Add G20/G21 and M149 support; refactor code_value functions

reidrankin:
(This PR is an update of #194.)

G20/G21 switch between inch and millimeter modes. Inch mode is common in generated G-code for CNC machines. M149 handles switching between Celcius and Kelvin (as well as Fahrenheit and Rankine, because why not?) units for temperature specification.

Both of these features can be included or excluded via defines in Configuratiuon.h; by default G20/G21 support is enabled, because I've seen plenty of CNC G-code that uses it, but M149 is not, as I have yet to see a slicer use Kelvin!

My implementation of these features involves breaking the family of code_value functions into functions that refer specifically to the values they request, be it linear distance, time, temerature, or simply a scalar value. The new functions are inlined where appropriate, and calls have been replaced with ones appropriate to the type of the value they are intending to retrieve. (This last optimization alone saves ~500 bytes by avoiding conversions to and from floating point when the only valid values are integers.)

Feel free to ask questions! I kinda got into this and just kept running, so there's more here than I initially expected to be.


reidrankin:
According to my very basic checks here, compiling with defaults for an Atmega 2560, both new features off saves 562 bytes over the current master branch through optimizations alone, turning inch mode support on adds 536 bytes, M149 support adds 288, and both together add 854.

clefranc:
@reidrankin I think these aren't new features. They are basic functionalities. They should not be switchable in configuration, they should be fully integrated to the main code.

That said, I have no decision-making power here, whether the *value* of your code is sufficient to be integrated into Marlin, please do not modify your code on the sole basis of my opinion.

AFAIK, your code looks very good!

Wackerbarth:
@reidrankin -- To some extent, I like your approach. However, I do have some concerns.
First, ˚K and ˚R are of no value. I know of no one who uses those units for anything other than EXTREMELY low temperatures or for theoretical calculations. Neither of those apply to our printer environment. Even suppressed from compilation, they just add obscuring "noise" to the code.

On the subject of distances, I'm not sure of the value in handling non-metric units because you are only approximating them.  In particular, at some point we have to translate those distances into angular "steps". Done properly, that conversion needs to scale by a rational number. Since we cannot be precise in both Metric and English units, only one will be "exact" and the other only approximated. And the units that can be expressed exactly is really a function of the mechanics. For example, if I have a threaded rod, toothed belt or pulley driving an actuator, the pitch of the thread or diameter will be either Metric or English and that determines which units can be expressed exactly. It is probably better to have the slicer know which units to use and quantize the motion rather than approximating it in the printer itself.

Now, the idea of processing the units on input and, since we cannot control temperature accurately, allowing temperatures in either ˚C or ˚F are worth considering.

But, before you spend too much effort, I will remind you that there will be some reconfiguration of the files very soon. You might want to wait for the reopening of the 1.2 development to resubmit those changes.

reidrankin:
Ah, but the conversions actually are exact! For some time now, an inch has
been defined as exactly 25.4mm, so we can do perfect conversions.
Furthermore, the F-to-C conversion is exact as well, since 0C is exactly
32F and 100C is exactly 212F. That gives a conversion factor of exactly
1.8. Check Wikipedia if you don't believe me!

I put in the define flags for space constraints only; I agree that inch
mode at least is core functionality.

The definition of M149 specifies Celsius and Kelvin units, so I'd argue
that Rankine is the only superfluous one, and it's cool. How could a unit
with such a good name be omitted?!? And how could you resist the bragging
rights? It's only an extra 5 bytes or so!

Wackerbarth:
No, the conversion that you are doing are not exact.

It is true that 10 inches is exactly 254mm. However, those are not the units that you are using.
Likewise, you are not using exact units for temperature. But, as I have noted, we can only approximate temperature in our hardware and that makes the roundoff irrelevant.

And, “bragging rights” isn’t sufficient reason to add non-useful code.

Also, it is not the intention of Marlin to be a general CNC machine. There is no value in implementing “features” that are not useful for our target audience.

reidrankin:
I run a CNC machine off of marlin. I have an extruder that I swap out for the router when I want to do 3d printing, and I'd prefer not to have to reflash every time! It's not like G20 and G21 are odd G-codes either; Marlin's basically the [only](http://reprap.org/wiki/G-code#G20:_Set_Units_to_Inches) firmware that *doesn't* support them, even among the 3d-pinter-focused crowd.

I still don't exactly understand what you mean by the units not being exact. Obviously there's inevitable floating-point rounding, but the conversion factor of 25.4 is right there on line 1004 of Marlin_main.cpp. Like I said, an inch has been [https://en.wikipedia.org/wiki/Inch](defined) as 25.4mm since 1959.

And OK, I'll take Rankine out, but you're an official party pooper.

Wackerbarth:
@reidrankin -- Perhaps you SHOULD try to understand what I am saying about units. The problem is that if you have a stepper motor that moves in increments of, for example 0.05mm, then if you are using inches instead, the roundoff WILL show up in the layer thickness.  Your slicer can compensate for that in ways that the printer never can because the slicer has control of the values sent to the printer and the printer must try to obey those instructions.

The same thing is true if I use a DHT11 to measure temperature and attempt to run my control loops in ˚F. The roundoff will mess up the calculated integrals and derivatives.  But, if we only ˚F for setting a desired target and for display, then our error is bounded by the roundoff.

As long as the user understands that these roundoff errors are occurring, and has the option to do the conversion before the values reach the printer, I have no objection to converting the inputs.

I fully expect that we will include these changes in our next iteration of the firmware.

As for being a "party pooper", I will admit "guilt". But Marlin isn't meant to be a Fraternity Party. Having useless features in our source actually costs us in that it makes the source more difficult to understand, and, therefore, maintain.

clefranc:
@reidrankin There is a new astyle check in Travis. All code submitted should now be computer formatted.

reidrankin:
OK, rebased against latest master, restyled, and this should fix the build error I introduced last commit (that'll teach me to compile with my own options off).

@Wackerbarth: I'm not sure where I've caused offense, but I'm sorry. I really am trying to understand what you're pointing out about units. I can certainly see that if you're trying to print a 1-inch box, say, with 0.5mm steppers, then your layers are going to have problems. That said, what units your G-code is in and what units your model itself is in are not necessarily the same. Most of my models are done using mm as a base unit, and usually using round numbers of them, but when I put those models though tools that output inch-mode G-code, I get instructions like "G1 X0.0393701". Those instructions are precise to a far greater degree than my steppers might be. My code would convert that to an offset of 1.0000005mm (because of 32-bit round-off), which is not perfect, but them's the breaks when you're using single-precision floats. Assuming that the direction of the round-off error is more-or-less random, it will take nearly a billion moves in the same axis to accumulate an error that would cause an error of a single step using a 0.5mm stepper using 16 microsteps. (For reference, a print of an Eiffel tower using default settings on Slic3r I just ran as a test had 40,000 moves in the X axis.)

I think a very similar problem to what you're envisioning would happen when trying to print models with dimensions that are not multiples of millimeters using (ordinary) millimeter-mode G-code and steppers. Looking at it that way, these errors are just something everyone has to live with, and is in fact living with already -- not something new introduced by multiplying by 25.4.

Finally, even if inch-mode did introduce quantifiable errors when printing models designed in millimeters, the blame lies more on the user who decided to feed the printer inch-mode G-code than the printer for the inevitable rounding errors. That said, the user might still make that choice; as I said, lots of legacy (and even state-of-the-art) special-purpose tools create inch-mode code either by default or exclusively. Maybe they aren't the tools you encounter all the time while doing average 3d printing operations, but they're out there, and so is their inch-mode code.

Wackerbarth:
> On Sep 21, 2015, at 12:36 AM, Reid Rankin <notifications@github.com> wrote:
> 
> OK, rebased against latest master, restyled, and this should fix the build error I introduced last commit (that'll teach me to compile with my own options off).
> 
That you could do so is really more the fault of our build environment for allowing you to attempt to do so.

> @Wackerbarth <https://github.com/Wackerbarth>: I'm not sure where I've caused offense, but I'm sorry.
> 
First, IMHO, you have not caused offense. We simply have a miss-communication which we need to resolve.
> I really am trying to understand what you're pointing out about units. I can certainly see that if you're trying to print a 1-inch box, say, with 0.5mm steppers, then your layers are going to have problems.
> 
Exactly.
> That said, what units your G-code is in and what units your model itself is in are not necessarily the same.
> 
Correct. But the important part is that you need to be doing the conversions to the units of the machine.

If I have a model in inches, and my machine also operates in inches, then I should not be converting to millimeters at all.
> Looking at it that way, these errors are just something everyone has to live with, and is in fact living with already -- not something new introduced by multiplying by 25.4.
> 
No. Because you cannot express 25.4 exactly, you are introducing additional error.

So, my concern with your code is that it has no provision to properly handle a machine which has inches as its fundamental unit.

reidrankin:
@Wackerbarth: I think I see what you're getting at now: The IEEE 754 single-precison representation of 25.4 is 0x41CB3333, which is really more like 25.399999618530273 (or so). That constant is off by the same amount each time, meaning the errors compound pathologically; a 0.5mm stepper using 16 microsteps per step will accumulate an error of a single microstep in only about 41k steps, which is definitely within the realm of possibility for a single print.

I suggest solving the problem by multiplying by 254 and then dividing by 10 instead. Those values can be expressed exactly, even in floating-point, which will mean the error accumulated by each operation will be dependent on the input, not the hardcoded constant. Instead of accumulating in the same direction each time, these errors should form a Bernoulli distribution, meaning the single-microstep-error in the situation described above will take 1.7 billion steps to occur instead.

reidrankin:
(Man, I give up on this rebasing--manually resolving the astyle conflicts was bad enough! Just let me know when you're done rearranging the entrails of the project and I'll do it all in one go.)

Wackerbarth:
@reidrankin -- I think I got all the files moved to their revised locations. The revised layout builds locally for me. However, I still have work to do on the Travis tests. They are failing miserably. So, you should be safe moving the files around. Except for some minor changes to the configuration files, I haven't changed the content.

Why did you "... manually resolving the astyle conflicts"? It's much easier to simply run astyle and let it do its thing.

I'll let you know when I get things back to working.

reidrankin:
Because I didn't notice it was using astyle until I committed and the style check failed! I thought you'd just very comprehensively restyled everything to remove all whitespace and semblance of readbility. Of course, git couldn't identify the right places to merge either, so by the bottom of Marlin_main.cpp I disliked you rather strongly.

And then clefranc mentioned astyle and I felt stupid.

thinkyhead:
Sorry this has been lingering for so long. Most attention has been going into fixing bugs and getting the "entrails of Marlin" to stop hemorrhaging so much. We thought it was a good idea to deal with all extant issues first and set aside this parallel initiative to reorganize the code until after 1.1.0 is released.

So, the plan is to bring over the `RCBugFix` code, but reorganize the file layout in the same manner as the `MarlinKimbra` fork. Then the plan is to go through all the issues here in MarlinDev and see which ones are already addressed. For issues like this one, I will take whatever time is needed to rebase it onto the new code and get it done properly for the latest code-base.

So, it's not forgotten, just on hold for the moment. I expect the final release of 1.1.0 to be ready by the end of April, and then we will jump back into this re-org with both feet.

boelle:

jbrazio:

Issue #162, opened 2015-09-11
FIX: Homing after MANUAL_BED_LEVELING issue.


spaelectronics:
I have been testing the latest MarlinDev (master) firmware, and I found an issue related to homing (G28) that only occurred after doing the "Manual / Mesh Bed Leveling" process.

**PROBLEM:**
When homing (G28), it would home to the endstops, then it would move the X & Y to 10 mm, which was displayed on the LCD correctly, but not in Repetier Host, which said the X & Y were at 0.

Secondly, the LCD stated that the Z-Axis was at 4mm, when it clearly was still at 0mm.

Long Story Short, after doing the bed leveling, G28 puts the hotend at 10,10 when it should be 0,0.

**Here is where the problem is in the code:**

**FILE: Marlin_main.cpp**

```
  // For manual leveling move back to 0,0
  #if ENABLED(MESH_BED_LEVELING)
    if (mbl_was_active) {
      current_position[X_AXIS] = mbl.get_x(0);
      current_position[Y_AXIS] = mbl.get_y(0);
      set_destination_to_current();
      feedrate = homing_feedrate[X_AXIS];
      line_to_destination();
      st_synchronize();
      current_position[Z_AXIS] = MESH_HOME_SEARCH_Z;
      sync_plan_position();
      mbl.active = 1;
      #if ENABLED(DEBUG_LEVELING_FEATURE)
        if (marlin_debug_flags & DEBUG_LEVELING) {
          print_xyz("mbl_was_active > current_position", current_position);
        }
      #endif
    }
  #endif
```

eboston:
Why?  If the user needs the position, they can just send the M114 
command.  No need to force it on everyone whether they need it or not.

drbogger:
Because when printing from Repetier Host, when it homes, it thinks its at 0 position, but when the print starts, it starts 10mm past where it should.

eboston:
So how does sending the position back when you home effect anything?

drbogger:
It lets repetier host know where it's at.

eboston:
but you already said that Repeiter host thinks it at 0,0 so all you would be doing is telling it what it thinks.

drbogger:
Your wrong. With manual_bed_leveling enabled, it homes, then immediately after it homes, it moves X & Y +10mm. But Repetier Host still thinks its at 0,0, when its actually at 10,10.

eboston:
[...]

PR #199, opened 2015-09-19
Move to Z=0 for G28 when using Manual Bed Leveling.

epatel:
• Move to Z=0 for G28 when using Manual Bed Leveling.
• Move up after sampling last point when doing Manual Bed Leveling
• Added my full name

See https://github.com/MarlinFirmware/MarlinDev/issues/162


epatel:
@drbogger Yes, if you home less than all axises the mesh bed levelling is not activated. I did not have time to investigate, solve and test that during this weekend. So at the moment turning it off (not turning it on after a partial G28) is the safe thing to do I think.

Marlin/ultralcd.cpp
```diff
@@ -468,6 +468,12 @@ static void lcd_tune_menu() {
   // Speed:
   //
   MENU_ITEM_EDIT(int3, MSG_SPEED, &feedrate_multiplier, 10, 999);
+#if ENABLED(MANUAL_BED_LEVELING)
+  //
+  // Bed Z:
+  //
+  MENU_ITEM_EDIT(float43, MSG_BED_Z, &mbl.z_offset, -1, 1);
+#endif
```
drbogger: Why is "Bed Z" located in the "Tune" Menu? Does this do anything specific "during" a print?

epatel: Yes, you can "tune" it while printing. i.e. when the brim is layed out and one notice it is squashed or too loose to stick. But be careful...

drbogger: Isn't that what the "Babystepping" feature is supposed to do?

epatel: Yes, you are right. Seems like that could be used for the same thing. Haven't seen it before, also it looks like it has been deprecated. I added this (+saving to EEPROM together with the rest of the mesh values) mostly because I had this problem that with different bed "textures" I'd need different nozzle bed distance for first layer. First I used the offset in Slic3r but felt like moving it to the printer was better (and have the value be stored).

drbogger: Hmm.. I'll test yours out and see how it works.  Have you tested it yet from the "Tune" menu?

epatel99: Yes, I have used it, and I am using it. The `0.001` is small yes, it is the standard menu item and I also think one should do very small changes when printing. I usually change to about `0.050` which can be made fairly easy on the display I have.

epatel99: The item seem to be used for some filament menu items. You could try to modify the `1000` on the line that has this `menu_edit_type(float, float43, ftostr43, 1000)`

drbogger: Well I'm not sure why, but I can't get it to work. I try to adjust it during a print, and the Z Axis does not move at all...
I'll see if I can figure out why, and report back.

drbogger: Okay, I can't seem to figure it out... any ideas?

drbogger: Okay, from what I can grasp from this, is that "Bed Z" is only used when homing. If I set "Bed Z" from the "Control > Motion" menu, then I do a "Prepare > Auto Home", it'll add/subtract the amount set in "Bed Z".

But with the "Tune > Bed Z" menu, it doesn't move the Z-Axis on the fly.

epatel: Please note that it will not change instantaneously. It changes similar to changes in speed, delayed because of the planning buffer. appr 10-12 lines I think.

drbogger: Oh, I see. I'll check it out again later. 
I wish there was a way to make it instantaneous.

thinkyhead:
Is there anything here that we haven't incorporated into the `RCBugFix` code yet? If not, we can close this issue. Code from `RCBugFix` will be getting migrated to `MarlinDev` (and reorganized) right after 1.1.0 release. If this PR contains changes that we want, I guess hold off for now and we can rebase it onto the latest code as soon as we get started on 1.2.x.

epatel:
@thinkyhead Think all these finally got in with this PR https://github.com/MarlinFirmware/Marlin/pull/3232

PR #200, opened 2015-09-19
Fix for LCD Contrast issues

eboston:
Removes the 0-63 range limit for contrast values.  Replaces the ENABLED
macro that was being used on defined values that were not 0 or 1.  This
is a temporary fix until the ENABLED/DISABLED macros can handle this
type.

### DIFF
--- a/Marlin/Configuration_adv.h
+++ b/Marlin/Configuration_adv.h
@@ -295,7 +295,7 @@
   // as SD_DETECT_PIN in your board's pins definitions.
   // This setting should be disabled unless you are using a push button, pulling the pin to ground.
   // Note: This is always disabled for ULTIPANEL (except ELB_FULL_GRAPHIC_CONTROLLER).
-  #define SD_DETECT_INVERTED
+  //#define SD_DETECT_INVERTED
 
   #define SD_FINISHED_STEPPERRELEASE true  //if sd support and the file is finished: disable steppers?
   #define SD_FINISHED_RELEASECOMMAND "M84 X Y Z E" // You might want to keep the z enabled so your bed stays in place.
--- a/Marlin/ultralcd.cpp
+++ b/Marlin/ultralcd.cpp
@@ -1163,13 +1163,9 @@ static void lcd_control_volumetric_menu() {
 #if ENABLED(HAS_LCD_CONTRAST)
 static void lcd_set_contrast() {
   if (encoderPosition != 0) {
-#if ENABLED(U8GLIB_LM6059_AF)
     lcd_contrast += encoderPosition;
-    lcd_contrast &= 0xFF;
-#else
-    lcd_contrast -= encoderPosition;
-    lcd_contrast &= 0x3F;
-#endif
+    lcd_contrast = constrain(lcd_contrast, LCD_CONTRAST_MIN, LCD_CONTRAST_MAX);
+
     encoderPosition = 0;
     lcdDrawUpdate = 1;
     u8g.setContrast(lcd_contrast);
@@ -1688,7 +1684,7 @@ void lcd_reset_alert_level() { lcd_status_message_level = 0; }
 
 #if ENABLED(HAS_LCD_CONTRAST)
 void lcd_setcontrast(uint8_t value) {
-  lcd_contrast = value & 0x3F;
+  lcd_contrast = value;
   u8g.setContrast(lcd_contrast);
 }
 #endif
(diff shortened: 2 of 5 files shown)


eboston:
Hmmm....  What did Travis fail on?

clefranc:
@eboston There is a new astyle check in Travis. All code submitted should now be computer formatted.

eboston:
@clefranc Interesting that Travis barfed on the first PR I did but not the second.  Both were based off the same base code and changes in neither PR changed the format.  If Travis is going to report a failure because of formatting, then it needs to say that is the reason for the failure, not some cryptic, generic, failure message.

eboston:
If running the astyle is going to be a requirement for submitting PRs, then why is there not a link to the file and instructions on doing so in the readme file or other prominently visible location?

Wackerbarth:
"If Travis is going to report a failure because of formatting, then it needs to say that" -- I agree. But you have a choice. You can either put up with the cryptic messages until such time as I get time to improve them (or, even better, submit a PR to improve them) or you can wait to submit anything.

As you were warned, we are in a transition. There are more important aspects that need to be changed first.

"but not the second." -- As far as I can tell, it did "barf" on both of them.

"then why is there not a link to the file ..." -- Because YOU didn't help improve the documentation.  :)

eboston:
@Wackerbarth I wasn't even aware the style change had been implemented.  I know it was being talked about but never saw a notice about it being in place.  As for the cryptic messages, how is anyone suppose to know that was the reason for the failure when there is nothing in the Travis logs to indicate why it failed?  If the failure notifications were not ready and in place, and more important aspects to be implemented first, this should not have been implemented yet.

As for barfing on both, no, only the first one.  I did something with git that caused the two PRs to be included together but it is still only reporting the failure on the first.

And why should it be the job of someone who didn't implemented a change to update documents with information they don't have?  It should be the responsibility of the person making those changes to document how it effects the system *BEFORE* they implement the change.  And how is someone other than the implementer suppose to know the documents need to be updated?

Wackerbarth:
Patience, Grasshopper.  The style change MUST be implemented immediately to keep people like yourself for continuing to push non-conforming changes into a system that isn't ready for them.

So, just quit trying until you see an announcement that we have completed the transition.

eboston:
You know what.  FUCK YOU!  I try to help and just get slammed when I hit the invisible hurdle you put in place and don't tell anyone about how to get over them.  So until you can get your act together, I will take your advice and not submit anything.  But I will take it further than that and *NEVER* submit anything again.  I have been slammed with every attempted I have made with submissions because it doesn't meet someone's personal style.  There are too many captains trying to steer this thing and I am getting sea sick from all the unannounced course changes.

I am going to close these two PRs and leave it up to you to fix the bugs yourself.  If you have more important things to do than to document major changes in the PR submission process, I am not going to waste my time trying to figure out how to get around them.

Marlin/ultralcd.cpp
```diff
@@ -1163,13 +1163,9 @@ static void lcd_control_volumetric_menu() {
 #if ENABLED(HAS_LCD_CONTRAST)
 static void lcd_set_contrast() {
   if (encoderPosition != 0) {
-#if ENABLED(U8GLIB_LM6059_AF)
     lcd_contrast += encoderPosition;
-    lcd_contrast &= 0xFF;
-#else
-    lcd_contrast -= encoderPosition;
```
thinkyhead: @eboston I'm just re-doing your feature for RCBugFix now. I am wondering about why the code did this in the first place. Was the encoder going in the opposite direction on non-`U8GLIB_LM6059_AF` controllers… or was this never actually needed?

Marlin/Configuration_adv.h
```diff
@@ -295,7 +295,7 @@
   // as SD_DETECT_PIN in your board's pins definitions.
   // This setting should be disabled unless you are using a push button, pulling the pin to ground.
   // Note: This is always disabled for ULTIPANEL (except ELB_FULL_GRAPHIC_CONTROLLER).
-  #define SD_DETECT_INVERTED
+  //#define SD_DETECT_INVERTED
```
thinkyhead: So far we've kept the default for this option _enabled_. It seems to be ok.

Marlin/ultralcd.cpp
eboston: All I remember about it was it was going the wrong way on my unit.  It could have been because I wired the connectors the opposite way as other displays.  It was just easier to fix in code this way than remaking the PCB.

Marlin/Configuration_adv.h
eboston: This is going to depend on the SD socket used.  The one I used in my prototype was for a full size SD Card.  The one I used on the board was for a micro SD Card.  The detect line were inverses of each other.  High for one when inserted.  Low on the other.  So if the maker of any of the other displays out there changes the type of socket they use, the user may have to change the setting.

thinkyhead: > the user may have to change the setting

That's okay, and exactly why it is here. So users can change it if they need to. This default works ok for most, apparently.

PR #201, opened 2015-09-19
Commented out SD_DECTECT_INVERTED

eboston:
The comments around this define states it should be disabled UNLESS your
are using....  But the defined was not disabled by default.  It is now.


clefranc:
@eboston Your correction should be applied (if needed) to all example files.

eboston:
Okay.  Applied to all the inconsistent Configuration_adv.h files.  Hopefully the new reorg of the files will remove this busy work necessary for these simple fixes.

Wackerbarth:
"Hopefully the new reorg of the files will remove this busy work necessary for these simple fixes."

That's the intention.

PR #222, opened 2015-09-30
Galician (gl) language.

rafacouto:
These are the language definitions for [Galician](https://en.wikipedia.org/wiki/Galician_language) language (ISO code "gl").


rafacouto:
I will send another PR on the right branch...

rafacouto:
See PR #224

PR #228, opened 2015-10-04
Fix compile bug with DEBUG_LEVELING_FEATURE

thinkyhead:
Same thing as MarlinFirmware/Marlin#2671 for the RCBugFix branch, only this is for the `MarlinDev/master` branch.

Is it required to make 2 PRs for each change now, or…? How do we propose to keep parity between `RCBugFix` and this branch? Keeping in mind that any bugs discovered in RCBugFix will surely apply to this branch, and vice-versa.


Wackerbarth:
@thinkyhead -- I see no reason to have the originator create separate PRs. A comment to the effect that a change also needs to be back/forward ported should be sufficient to allow the integrator to post it to all branches. When the two branches start to have significant separation, it may be desirable to alter the procedure to allow separate testing. However, it is probably easier to do that internally since we can bypass the rebase step and post the ported version ready to merge.

PR #230, opened 2015-10-05
Fix: G-Code parser tripping on "N"

thinkyhead:
Problem: The G-Code parser complains about the Line Number if it sees the letter `N` anywhere in the command. Thus, when sending commands (from the console) commands like `M23` and `M33` may only use lowercase in their dos paths. Even the `M117` command must not use the letter `N` in the printed string or it trips the error.

Solution: Instead of seeking for the `N` (line-number) parameter in the whole command, only look for it at the start of the command. This should be compliant with all hosts and other software generating G-Code with line numbers at this time.


Wackerbarth:
Incorporated

PR #235, opened 2015-10-08
Limit the display selectable move_to_x/y coordinates for deltas to a …

AnHardt:
…reachable area Va.

This is version a. (devil)
Here i don't like calling delta_clip() (calculating a sqrt() ) twice with the same parameters.

The use of the constant is to be discussed.
~~a) if DELTA_PRINTABLE_RADIUS is adjustable somewhere menu/g-code/EEPROM we have to use the variable instead.~~ checked - no.
b) if it's a real constant - What name?


thinkyhead:
Very nice fix.

__Pro Tip:__ For little tweaks to commits, it's handy to use `git commit --amend` or `git rebase HEAD~2` (and apply "fixup" to commits that are just… fixups). Then you can use `git push -f` and get a nice clean PR with no broken commits (and nothing to squash later in `dev`).

AnHardt:
@Wackerbarth 
If you delete a branch all open PRs to this are closed.

thinkyhead:
Was this PR re-done? Did I recently merge it? :smile: I feel like I did….

AnHardt:
You included it in your #3006

thinkyhead:
> You included one, of this kind…

Oh yeah, I see the solution to prevent bad movements from the LCD, and it seems good enough. I see that it will do things like prevent moving X very far if Y is far away from the center. In the long run, it might be good to allow X or Y to move in the full range, but have the other axis move as-needed towards the center to make room for the new X or Y position being set.

PR #236, opened 2015-10-08
Limit the display selectable move_to_x/y coordinates for deltas to a …

AnHardt:
…reachable area. Vb

This is version b. (Lucifer)
Here i don't like the additional delta_clip_duplicator variable.

Take what ever you want.

The use of the constant is to be discussed.
~~a) if DELTA_PRINTABLE_RADIUS is adjustable somewhere menu/g-code/EEPROM we have to use the variable instead.~~ Checked - no.
b) if it's a real constant - What name?


ultralcd.cpp
```diff
@@ -813,8 +813,16 @@ static void _lcd_move(const char* name, AxisEnum axis, int min, int max) {
   if (lcdDrawUpdate) lcd_implementation_drawedit(name, ftostr31(current_position[axis]));
   if (LCD_CLICKED) lcd_goto_menu(lcd_move_menu_axis);
 }
-static void lcd_move_x() { _lcd_move(PSTR(MSG_MOVE_X), X_AXIS, X_MIN_POS, X_MAX_POS); }
-static void lcd_move_y() { _lcd_move(PSTR(MSG_MOVE_Y), Y_AXIS, Y_MIN_POS, Y_MAX_POS); }
+#if ENABLED(DELTA)
+  static float delta_clip_radius_2 =  DELTA_PRINTABLE_RADIUS * DELTA_PRINTABLE_RADIUS;
+  static int delta_clip_duplicator;
+  static int delta_clip( float a ) { delta_clip_duplicator = sqrt(delta_clip_radius_2 - a*a); return delta_clip_duplicator}
```
thinkyhead: Oddly the compiler isn't complaining here, but I note that `return delta_clip_duplicator` doesn't have a semicolon on the end.

ultralcd.cpp
```diff
@@ -813,8 +813,16 @@ static void _lcd_move(const char* name, AxisEnum axis, int min, int max) {
...
+  static void lcd_move_x() { _lcd_move(PSTR(MSG_MOVE_X), X_AXIS, max(X_MIN_POS, -delta_clip(current_position[Y_AXIS])), min(X_MAX_POS, delta_clip_duplicator)); }
```
thinkyhead: Actually, I think it's cleaner (and less brittle) just to do it this way:
```cpp
static void lcd_move_x() { int clip = delta_clip(current_position[Y_AXIS]); _lcd_move(PSTR(MSG_MOVE_X), X_AXIS, max(X_MIN_POS, -clip), min(X_MAX_POS, clip)); }
```
That way we don't have to foul up global namespace with another very temporary value, or rely on "this thing (`delta_clip_duplicator`) was set by that function (`delta_clip`) that was called earlier (because `max` called it)." Instead just explicitly set an auto var while only needing to call `delta_clip` once (which I presume was the original caveat).

thinkyhead: On another note, shouldn't it actually be `delta_clip(current_position[X_AXIS])` for `lcd_move_x` and `delta_clip(current_position[Y_AXIS])` for `lcd_move_y`?

AnHardt:
@thinkyhead @clefranc 
For a square bed the limit for x is X_MIN_POS, X_MAX_POS, foy y Y_MIN_POS, X_MAX_POS.
That's what we had before.
For a round bed we have to solve (x*x + y*y = R*R). Where R is ether the radius of the bed or the radius the delta machine can work in. Here x depends on y and y on x.
If we have a delta, what is further limited by a (small) square bed, we need a square with rounded corners to describe the allowed area. That's what i tried to express in this variation of the PR.

But if R is calculated from the beds max/min /x/y we always have a circle fitting on the bed but loosing a lot of printable area. That's why i asked for an alternative to DELTA_PRINTABLE_RADIUS or an alternative calculation for that

See more indo at https://github.com/MarlinFirmware/Marlin/issues/2679

AnHardt:
Repaced by #270

PR #237, opened 2015-10-09
Added EEPROM functions to store total number of completed prints

iashwinprasada:
using the EEPROM of arduino we have successfully tested the function to count the number of completed prints...

functions reside in the ConfigurationStore.cpp and is called from CardReader.cpp (CardReader::printingHasFinished) after each print

also added LCD menu funtions which will read from the EEPROM and display the total number of prints on LCD!

also added M codes:
M504: to retrieve the total number of prints from serial
M505: Just to call increment function for testing ONLY!
M1000: to reset the counter

i've commented  "Total_Prints"  to all the new code

Best,
EZ3


AnHardt:
Superseded by #249

PR #241, opened 2015-10-12
Watch E temperatures when changed from LCD

thinkyhead:
- Addresses longstanding bug MarlinFirmware/Marlin#2315.
- Reduce binary size by removing redundant code

### DIFF
--- a/ultralcd.cpp
+++ b/ultralcd.cpp
@@ -475,11 +475,71 @@ void lcd_set_home_offsets() {
 #endif //BABYSTEPPING
 
 /**
+ * Watch temperature callbacks
+ */
+#if TEMP_SENSOR_0 != 0
+  void watch_temp_callback_E0() { start_watching_heater(0); }
+#endif
+#if EXTRUDERS > 1 && TEMP_SENSOR_1 != 0
+  void watch_temp_callback_E1() { start_watching_heater(1); }
+  #if EXTRUDERS > 2 && TEMP_SENSOR_2 != 0
+    void watch_temp_callback_E2() { start_watching_heater(2); }
+    #if EXTRUDERS > 3 && TEMP_SENSOR_3 != 0
+      void watch_temp_callback_E3() { start_watching_heater(3); }
+    #endif // EXTRUDERS > 3
+  #endif // EXTRUDERS > 2
+#endif // EXTRUDERS > 1
+
+/**
+ * Items shared between Tune and Temperature menus
+ */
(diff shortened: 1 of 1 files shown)


ultralcd.cpp
```diff
@@ -475,53 +475,51 @@ void lcd_set_home_offsets() {
...
+#endif
+#if EXTRUDERS > 1 && TEMP_SENSOR_1 != 0
+  void watch_temp_callback_E1() { start_watching_heater(1); }
+  #if EXTRUDERS > 2 && TEMP_SENSOR_2 != 0
+    void watch_temp_callback_E2() { start_watching_heater(2); }
+    #if EXTRUDERS > 3 && TEMP_SENSOR_3 != 0
+      void watch_temp_callback_E3() { start_watching_heater(3); }
+    #endif // EXTRUDERS > 3
+  #endif // EXTRUDERS > 2
+#endif // EXTRUDERS > 1
```
AnHardt: Gosh! That simple! And i broke my head to avoid multiple functions an wanted to squeeze all in on - what failed.
Thanks.

AnHardt:
For some unknown reason this slipped thru Travis.
When i compile i get:
```
sketch\ultralcd.cpp: In function 'void nozzle_bed_fan_menu_items()':

ultralcd.cpp:185: error: '_menuItemNr' was not declared in this scope

       if (_menuItemNr == _lineNr) { \
...
```
and more errors about not declared variables normally declared in  START_MENU();

AnHardt:
I guess `nozzle_bed_fan_menu_items()` can not work as a function because the menu-macros rely on the function-local variables declared in START_MENU(). Maybe it can work as a macro but that would not save space.

AnHardt:
When the calls of `nozzle_bed_fan_menu_items()` are replaced with the contend of `nozzle_bed_fan_menu_items()`, the fix for MarlinFirmware/Marlin#2315 works perfectly.

thinkyhead:
Ah, right. There are local vars that the code in that function needs so it can hook into the menu system. The solution is actually pretty nice. Just pass the local vars by reference using the names that the macros expect.
```cpp
static void nozzle_bed_fan_menu_items(uint8_t &encoderLine, uint8_t &_lineNr, bool &wasClicked) { ... }
...
nozzle_bed_fan_menu_items(encoderLine, _lineNr, wasClicked);
```

AnHardt:
Normally cal by reference is the other way around.
```
static void nozzle_bed_fan_menu_items(uint8_t *encoderLine, uint8_t *_lineNr, uint8_t *_drawLineNr, uint8_t *_menuItemNr, bool *wasClicked, bool *itemSelected) {
 { ... }
...
nozzle_bed_fan_menu_items(&encoderLine, &_lineNr, &_drawLineNr, &_menuItemNr, &wasClicked, &itemSelected);
```

Issue #195, opened 2015-09-17
Code refinement only


justmyopinion:
In temperature. cpp
watchdog_reset should be made conditional.

thinkyhead:
You'll note that the original developers of Marlin chose not to make it conditional in `temperature.cpp` but instead to define it as an empty function in `watchdog.h` when `WATCHDOG` isn't enabled. I agree that making it conditional so that `watchdog.h` isn't even included is cleaner.

PR #243, opened 2015-10-12
Simplify watchdog slightly

thinkyhead:
Addressing suggestion #195
- Move reset_watchdog to an inline in the header
- Remove extra code to define null functions when `USE_WATCHDOG` is disabled
- Make watchdog code conditional as-usual in `temperature.cpp`
- Watchdog doesn't need `ultralcd.h` AFAICT


thinkyhead:
Rebased and looking good…

PR #249, opened 2015-10-22
Print Counter

RicardoGA:
Adds the function to store the number of prints in the EEPROM . Is based
in the code of @iashwinprasada with a few changes to make it work
independently if you have or not enable the EEPROM Settings.

Also i did a few changes in the way of implement this function with the
lcd, now is ready to add this feature in different languages currently
in (English and Spanish) need to translate the new menu
(MSG_PRINT_COUNTER, MSG_PRINTS).

The new Gcodes are

M504 - Retrieve the total number of prints
M505 - Increase the print counter by 1
M506 - Reset the print counter

And you can check the number of prins with the lcd in:

Menu\Control\Print Counter


AnHardt:
This new feature should not be compiled in every time - it needs to be configurable.
The ` int i=100;` is in the code several times. The '100' should be replaced by a named (#define) value. 
Why did you use '100'? Or asked the other way around, why is EEPROM_OFFSET now '150'? Can this be calculated?

iashwinprasada:
I've moved the EEPROM_OFFSET to 150 because I wasn't sure how much memory
the settings were taking, didn't wanna mess up that space! Moving it to 150
and using memory from 100 for print count seemed alot safer... But I'm sure
you guys would've a more elegant solution?

Wackerbarth:
I have an alternate approach to suggest.  The current EEPROM stores a fixed struct{} (which gets revised from time to time). Instead, we could replace that struct with a key/value dictionary
The dictionary has two advantages:
• The actual order in the memory is not fixed. Additions and deletions do not affect other parameters.
• In addition to fixed sized values, we can store variable length strings. This would be very useful if we start replacing some of the more complex code with the G-code (or equivalent) representation of the desired motion.

Roxy-3DPrintBoard:
> M504 - Retrieve the total number of prints
> M505 - Increase the print counter by 1
> M506 - Reset the print counter

Ok...  Ok...   It's fine...  But really...   Going forward we shouldn't grab lots of M & G codes!!!

With that said, I'm thinking the Marlin Team should Reserve (Grab!!!)  G24, G25, G26 and G27 right now.   It might make sense to use G27 for as Bed Leveling.  And I have good ideas for G26 all the way down to G24.

Roxy-3DPrintBoard:
@RicardoGA  May I ask you to consider just using one MCode and having parameters that control the behaviour?   It is going to be easier for us to support and maintain this code if everybody knows M505 controls the Print Counter.   Maybe something like:

M505 G     Gets the current Print Count.    
M505 C      Clears the Print Count.   
M505 A      indicates Another Print has started.

And who knows???   Maybe:

M505 F     indicates a Print Finished.   

If we have all of the functionality in one command People will remember the numbers that are important to them.  And it lets us extend things easier to have more functionality like keeping tract of both started and finished prints.

iashwinprasada:
M505 G & M505 C sounds better, but I didn't get the A & F parameters...Why
will that be required here?

Roxy-3DPrintBoard:
> M505 G & M505 C sounds better, but I didn't get the A & F parameters...Why
will that be required here?

I've only skimmed the code.   But here is what I was thinking:

>M504 - Retrieve the total number of prints
>M505 - Increase the print counter by 1
>M506 - Reset the print counter

If you were willing to consider a change, M505 with no parameter or M505 G    would be the same as M504 currently is and 'Gets' the total number of prints so far.

M505 S would be equivalent to your M505 and indicate another print has Started.
M505 F was an addition just to show that this lets us expand the code in the future (or now) to keep tract of how many prints actually finished.  I don't know about you, but I start 5x the number of prints compared to the number that actually finish.  So this might be an interesting number to have in addition to the M505 S number.

And of course, you need a way to do your M506 to Clear the counters so for the sake of this discussion, M505 C would do that.

RicardoGA:
Hello sorry for the late response I was a little busy. @Roxy-3DPrintBoard I think its a great idea change the use of differents M codes to parameters. I'm going to change that and also add the option to enable and disable in the configuration as @AnHardt request

Nandox7:
Don't want to spoil the party but... why should this handled by the FW?
This would make more sense to be handled by an external app, like Octoprint does.

RicardoGA:
@Nandox7 i think it's useful that we can keep the track of printer usage independently of the external app we're using to print.
And it's a better way to keep a record for different purposes. ex: printer maintenance.

Roxy-3DPrintBoard:
> I'm going to change that and also add the option to enable and disable in the configuration as @AnHardt request

Thank You!  And please join the discussion that @Wackerbarth wants to have about a 'Registry' (my words... He might object to that thinking!) for important values stored in the EEPROM.

> Don't want to spoil the party but... why should this handled by the FW?
> This would make more sense to be handled by an external app, like Octoprint does.

Sure!  If we had control of each and every Client and everybody agreed what they needed and how they wanted it to be done...  That might be better.   But this is a feature that some people want now and it can be turned off by anybody that doesn't want it.  Later when PronterFace and Octoprint add their own version of it people will have a choice of which version (if any) they want.

That is the beauty of Open Source.   Things can move in which ever way the community wants it to move.

iashwinprasada:
@Roxy-3DPrintBoard yeah! Their can be a separate counter to store the total number of prints stopped in the middle..that can be done by incrementing a variable whenever we pass the "stop print" 

So three counters?
Total prints started
Total prints completed
Total prints stopped

By doing this We can also look at the success to failure ratio!

Wackerbarth:
> On Oct 27, 2015, at 11:27 AM, Roxy-3DPrintBoard <notifications@github.com> wrote:
> Thank You! And please join the discussion that @Wackerbarth <https://github.com/Wackerbarth> wants to have about a 'Registry' (my words... He might object to that thinking!) for important values stored in the EEPROM.
> 

Thinking, no. Wording, yes. I have an aversion to the use of a term associated with a particular vendor’s implementation of a concept when the objective is anything other than intentional compatibility with their code. It’s something like the distinction between generic and brand-name drugs.  In this case, the term that you are using was neither the first, nor the most common, use of the programming concept.

However, what I am suggesting would function within Marlin much as a ‘Registry' serves a similar purpose within that particular well know brand of computer operating systems.

Roxy-3DPrintBoard:
> @Roxy-3DPrintBoard yeah! Their can be a separate counter to store the total number of prints stopped in the middle..that can be done by incrementing a variable whenever we pass the "stop print"

>So three counters?
>Total prints started
>Total prints completed
>Total prints stopped

Can you help me understand this?   When I Stop a print, I just press Pause in PronterFace.   And I do what ever takes to clean up the bed and get things right.   And then I press Re-Start on PronterFace.

I don't think Marlin actually knows I stopped the print.    But presumably the GCode that Slic3r put at the end of the file never got sent and no command ever happened to say the print finished so that counter never got bumped.   However, when I press Re-Start that GCode that Slic3r put at the start of the file does happen again and bumps the counter.

How does this third counter get bumped when I stop a print?

KiteLab:
You dom't need a third counter for the unfinished (stopped) prints.
The difference between the regularly finished and the started prints is the number of uncompleted prints.

iashwinprasada:
Damn, you're right. All we need is to add two gcodes at the start and end of each file... One to bump the total prints started and the other to bump the completed ones! Sorry for confusing, this should be pretty simple to implement!

RicardoGA:
@Wackerbarth do you have an example of how do you want to handel the counter location? So i can make the changes to meet that way to handle the location

thinkyhead:
@RicardoGA The smallest EEPROM on any Arduino board is 512 bytes. Currently the total storage used by the `M500` command is 221 bytes. So there is still plenty of overhead if you want to store some persistent values where they won't get overwritten. A `long` or `float` takes up 4 bytes, so I suggest starting at position 508 and counting backward.

Code within `configuration_store.cpp` can use `EEPROM_WRITE_VAR(508, myCounter)` and `EEPROM_READ_VAR(508, myCounter)`. So you could add something like these to `configuration_store.cpp` and expose them as public methods in the header:
```cpp
void Config_SetPrintCounter(long c) { EEPROM_WRITE_VAR(508, c); }
long Config_ReadPrintCounter() { long c; EEPROM_READ_VAR(508, c); return c; }
```
The only down-side is that EEPROM has limited write cycles, so writing it on every print might shorten its life slightly.

Roxy-3DPrintBoard:
> The only down-side is that EEPROM has limited write cycles, so writing it on every print might shorten its life slightly.

Yes, this is true.   But at: http://www.farnell.com/datasheets/1693899.pdf  it quotes the minimum number of cycles at 100,000 for the EEPROM.   I doubt any printer is going to get near that number!    :)

RicardoGA:
@thinkyhead changing the count offset its a good idea, i just restore the eeprom offset and version and change the count offset to 508

thinkyhead:
Sorry for the very long wait! This feature is still appreciated, but we've been focusing all our attention on the 1.1.0 release before we get back to working on the 1.2.x code reorganization. As soon as we get 1.1.0 released, we will be reorganizing the code in the manner of MarlinKimbra. Once we get through that process I will see about rebasing this PR to the updated code.

boelle:

jbrazio:

PR #250, opened 2015-10-22
Babystep_Multiplicator_Fix

RicardoGA:
Restore the Babystep Multiplicator functionality and now it's available
for X/Y/Z

### DIFF
--- a/configurations/deprecated_configurations/Felix/Configuration_adv.h
+++ b/configurations/deprecated_configurations/Felix/Configuration_adv.h
@@ -371,7 +371,7 @@
 #if ENABLED(BABYSTEPPING)
   #define BABYSTEP_XY  //not only z, but also XY in the menu. more clutter, more functions
   #define BABYSTEP_INVERT_Z false  //true for inverse movements in Z
-  #define BABYSTEP_Z_MULTIPLICATOR 2 //faster z movements
+  #define BABYSTEP_MULTIPLICATOR 2 //faster movements
 #endif
 
 // @section extruder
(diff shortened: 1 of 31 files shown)


AnHardt:
Indeed BABYSTEP_Z_MULTIPLICATOR is not used anywhere in the code.

steps/mm can be considerable different for x,y and z. Does it make sense to have different multiplicators?

Wackerbarth:
I would like a clear explanation of just what baby stepping is attempting to accomplish.

Considering only the most basic case, you have three axes which are colinear. They might be designated as the Sample-axis, measured in mm, the Bed-axis, also measured in mm, and the Actuator-axis, perhaps measured in steps.

During printing, when the nozzle moves, it moves a proportionate distance as measured on any of these axes.  Does baby stepping also cause such motion? Or, does baby stepping move the origin of one axis relative to the others?

RicardoGA:
@AnHardt Thats right BABYSTEP_Z_MULTIPLICATOR is not used in the current code. In previous version the BABYSTEP_Z_MULTIPLICATOR was used with the code of z axis, but at some point the code change to controll all the axis X/Y/Z instead of continue independently and the multiplicator was removed.

Now i don't think it´s necesary a diferent multiplicators  for all the axis and the reason  is the babystep funtion in X/Y is barely used and with z axis is used to compensate the z height usually at the start of the print.

ex: when i change the my printer's nozzle  from .4 to .25 and i don't have time to calibrate the new height i  remove 1 mm to the nozzle offset and the when i print i ajust the z heigh with babysteping .

@Wackerbarth as far as i know, the baby stepping let you  move the axis in real time independently of endstop, but in steps that's why the movement is not proportional in all axes because don't take the steps per mm to make the movements.

AnHardt:
Also solving https://github.com/MarlinFirmware/Marlin/issues/2742 when set negative.
https://github.com/MarlinFirmware/Marlin/issues/2742 and #267 are not needed.

Issue #248, opened 2015-10-20
Marlin Main_cpp missing bracket


sebastian-dev:
in Marlin_main.cpp there is an missing bracket near line 7158

when you use the option: FAST_PWM_FAN

          TCCR5B |= val;
          break;
      #endif
  }
  } <----- this is missing

#endif // FAST_PWM_FAN

AnHardt:
Solved by merging https://github.com/MarlinFirmware/MarlinDev/pull/251

PR #252, opened 2015-10-26
Missing bracket issue #248

AnHardt:
Add the Missing bracket of FAST_PWM_FAN

Like https://github.com/MarlinFirmware/MarlinDev/pull/251
with corrected indendation.


AnHardt:
@Wackerbarth 
You have been a moment to fast.
Don't merge both (#251 and #252). Otherwise we will have onr '}' to much.

Wackerbarth:
@AnHardt - Yes, I recognize that. I have also offered a test configuration that, when we figure out how to best use it, will help us catch us should we make the mistake of doing both.

For the permanent commit stream, I will retain the version that has the better formatting.

Wackerbarth:
This has been incorporated

PR #253, opened 2015-10-26
Test for issue #248

Wackerbarth:
Perhaps we need to be developing test configurations that, at least, verify that the affected option sets compile correctly.

I do not recommend that these tests be run with every submission.
However, we might want to keep them around and run some extended set occasionally.

### DIFF
--- a/ReleaseNotes.md
+++ b/ReleaseNotes.md
@@ -1,0 +1,2 @@
+Marlin is released under the Open Source GNU GENERAL PUBLIC LICENSE, Version 3.
+See [LICENSE](/LICENSE) for details.
--- a/configurations/issues/FAST_PWM_FAN.h
+++ b/configurations/issues/FAST_PWM_FAN.h
@@ -1,0 +1,5 @@
+// Unbalanced brackets in Marlin_main.cpp when FAST_PWM_FAN is selected
+
+#define FAST_PWM_FAN
+
+#include "configurations/transitional_default_configurations/ramps/Configuration.h"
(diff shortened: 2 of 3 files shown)


AnHardt:
@Wackerbarth 
Good idea.

AnHardt:
How about naming the file `FAST_PWM_FAN.h` instead of  `Issue_248.h`?

All the filenames in the travis directory are not very 'talking'. 

Names could describe the feature the test. Its not necessary to list all the changed defines. Usually there is on feature we want to test and other defines to make that possible or are associated with this.

````
.../MarlinDev/configurations/travis
$ ls
AUTO_BED_LEVELING_FEATURE.h  RAMPS_13_EFB_default.h
COREXY.h                     RA_CONTROL_PANEL.h
COREXZ.h                     REPRAP_DISCOUNT_FULL_GRAPHIC_SMART_CONTROLLER.h
EEPROM.h                     REPRAP_DISCOUNT_SMART_CONTROLLER.h
FILAMENT_LCD_DISPLAY.h       SERVOS.h
FILAMENT_SENSOR.h            TEMP_SENSOR_BED.h
G3D_PANEL.h                  ULTIMAKERCONTROLLER.h
LCD_I2C_PANELOLU2.h          ULTRA_LCD.h
LCD_I2C_SAINSMART_YWROBOT.h  Z_DUAL_ENDSTOPS.h
LCD_I2C_VIKI.h               dual_extruder.h
MAKRPANEL.h                  tripel_extruder.h
PIDTEMPBED.h
````

Wackerbarth:
I think this is a reasonable proposal. The configuration files can really have any name that we like.
However, I would like to keep all of the Travis tests together (unless they are simply the standard build for some particular printer design)

PR #260, opened 2015-11-01
Removing Button updates fron ISR file

AnHardt:
lcd_buttons_update is done frequently (more often than in this context) in lcd_update so i do not see the need to have this burden in ISR routine also. i have not met any consequencies of removing all updates.

Originally by @justmyopinion in https://github.com/MarlinFirmware/Marlin/pull/2720

@Wackerbarth, @Roxy-3DPrintBoard  you seemed to be interested in this over here in Dev


justmyopinion:
This request has ben closed by owner and is no longer a valid request so do not capture my requests without permission

Roxy-3DPrintBoard:
> This request has ben closed by owner and is no longer a valid request so do not capture my requests without permission.

The topic is of interest to several of the developers.   It should be resolved.

@AnHardt   I agree.  I'm not sure other people would agree, but my thinking is even if the Button response becomes a little bit sluggish, that is OK because it means we are not spending lots of CPU cycles looking for button presses that hardly ever happen.

AnHardt:
@justmyopinion 
Hmm. I'm Sorry. 
You contributed to a open source project under GPL.
It's still available under https://github.com/justmyopinion/Marlin

But if you really don't wont it i'll remove it again.

justmyopinion:
Anhardt
My contribution was targeted 1.1 and not destined for the Wackers personally project so maybe you should have asked the owner before you sent my requests along the line so please regard all this a mistake and remove closed requests from marlinDev.


justmyopinion
Den 1. nov. 2015 kl. 23.41.26 +01.00, skrev AnHardt <notifications@github.com>:

AnHardt:
@justmyopinion 
I'm sorry.
Even i  currently don't feel very comfortable on the Dev branch. But for bringing in new features (or improvements needing a bit more testing (because i do not understand them completely)) i see no other way - at the moment.

Thanks again for your good contributions. I'll pick up some of your ideas soon.

thinkyhead:
It's a lot of extra work to keep the 1.1 RC patches synchronized with the dev branch, but I think it's a good idea to do so. It just makes things easier going forward. I have so far been unable to make any sense of the new MarlinDev layout and configuration, but I figure if I lay low for a few months maybe it will sort itself out and get proper documentation...?

Wackerbarth:
Yes, documentation is an issue. Unfortunately, I don't know where to put it. I don't want to confuse the 1.1 users with details that don't apply to 1.1. But the 1.2 users need those details.

Since there are some who advocate keeping the documentation in a wiki (which I don't oppose if we can provide write access to all of the major contributors), I'm looking to them to propose a way to keep the various versions.  Otherwise, my only suggestion is that we go back to putting the documentation in the Documentation folder (which automatically has a version for each branch). If put in a documentation folder, the format of the documentation (txt, md, html, etc.) is still to be ecided.

However, I definitely think we need to do it one way or another. Having documentation spread about in multiple locations leads to problems for both the users and the maintainers.

Roxy-3DPrintBoard:
How about we all agree we will just put the documentation in a separate
folder.  If somebody has the motivation and energy to get it moved to a
Wiki we can have the discussion at that point about removing the embedded
folder.

My thoughts are it doesn't hurt to have the documentation in two places.
But we definitely need the documentation in at least one place.

Roxy-3DPrintBoard:
> It's a lot of extra work to keep the 1.1 RC patches synchronized with the dev branch, but I think it's a good idea to do so. It just makes things easier going forward. 

I think it is a good idea to do this!   And there are a number of previously prepared patches that you have done that have not been accepted yet.  We certainly need to get the bug fix ones done.  

>I have so far been unable to make any sense of the new MarlinDev layout and configuration, but I figure if I lay low for a few months maybe it will sort itself out and get proper documentation...?

I haven't been using any of the MarlinDev stuff.  I've been trying to get RC2 up on my printer.   I'm sure my thoughts will need to be corrected on this topic.  I thought @Wackerbarth said things needed to be stable long enough to bring the library format development branch up and get it cooking.  Maybe I'm wrong, but it seems like the new format isn't coming up quickly?

If that is true, maybe we all need to have discussion and decide whether we push forward on the library idea or if we back off.   Because we do need to let people make changes to the development branch.  (And cross over bug fixes!!!)  I do know part of the problem is @Wackerbarth is travelling right now and doesn't have a lot of time to be on his computer helping with this.

PR #261, opened 2015-11-01
Some updates for smoother printing at high speeds

AnHardt:
115200 is adequate as default and higher baudrates gives trouble due to stepper ISR limits
changes in cmd buffer handling for more smoothness
removing some hack from stepper ISR.

Originally by @justmyopinion in https://github.com/MarlinFirmware/Marlin/pull/2719

A bit cherry picked to separate from #260 and https://github.com/MarlinFirmware/Marlin/pull/2691


Roxy-3DPrintBoard:
@AnHardt    Can you help me understand something?   The Serial Baud Rate is kind of meaningless if it is being transmitted (or received) across USB, right?   Is there some timer or something that this affects that controls how often a byte is sent across USB?   

But anyway...  Given how short the GCode commands are, if this really was a Serial USART, 115200 baud would be blasting commands so fast at the printer that there isn't a hope it could keep up.   If this gets us smoother printer, we want to bump this down.

AnHardt:
@Roxy-3DPrintBoard 
As long the transmission is on the USB lines the  Serial Baud Rate is meaningless.
But for example at a Arduino Mega the USB line ends at a USB-IC (in this case ATmega16U2 ), from there to the Atmel 2560 we have a real serial line again.

I have no doubt that 115200 baud is fast enough. My only objection is, that we'll have a lot of support questions "I can't connect ..." - at least in the first time.

AnHardt:
@Roxy-3DPrintBoard 
Thinking about it. On fast moving machines without microstepping and with extremely short moves there may be a problem. But that's not our typical use. If one wants a faster baud rate he still can go up to 250000.

Roxy-3DPrintBoard:
@AnHardt   So I think what you are saying is inside the AVR chip, there is both a USB interface and hiding behind that, inside the chip is a UART.   And that UART receives the characters from the USB and then behaves as any normal UART would and interrupts when it gets a character?   Is my understanding correct?

I agree that if some super fast moving machine sees a little bit of stuttering, it isn't the end of the world.  They can bump the baud rate up get pounded by interrupts.    Right now, we need to be conserving CPU cycles for the main stream users.

AnHardt:
@Roxy-3DPrintBoard 
There are several kinds of Atmel chips out there. Some have integrated USB interfaces, like the AT90USB1286 on the Printrboard, some need extra USB chips (FT232RL, ATmega16U2, ... ), like the ATmega2560 on the Arduino Megas.

PR #264, opened 2015-11-03
Fix DEBUG_LEVELING bug in Marlin_main.cpp

thinkyhead:
See MarlinFirmware/Marlin#2671


Roxy-3DPrintBoard:
I used your debug stuff again today!

PR #267, opened 2015-11-04
Reimplement BABYSTEP_INVERT_Z

AnHardt:
Fixing https://github.com/MarlinFirmware/Marlin/issues/2742


AnHardt:
Probably there are a lot of possible solutions. You can use the ENABLED-macro but you do not have to. We already have:
```
 #if ENABLED(BABYSTEPPING)
 
   static void _lcd_babystep(int axis, const char* msg) {
     if (encoderPosition != 0) {
-      babystepsTodo[axis] += BABYSTEP_MULTIPLICATOR * (int)encoderPosition;
+      babystepsTodo[axis] += BABYSTEP_MULTIPLICATOR * (int)encoderPosition * (BABYSTEP_INVERT_Z && axis == Z_AXIS) ? -1, 1;
       encoderPosition = 0;
       lcdDrawUpdate = 1;
     }
```
and in all the configurations we have:
```
#if ENABLED(BABYSTEPPING)
  #define BABYSTEP_XY  //not only z, but also XY in the menu. more clutter, more functions
  #define BABYSTEP_INVERT_Z false  //true for inverse movements in Z
  #define BABYSTEP_Z_MULTIPLICATOR 2 //faster z movements
#endif
```
` BABYSTEP_INVERT_Z` exists when `BABYSTEPPING` exists. For `BABYSTEPPING` we tested before.

If i had not tried to repair it, but wanted to invent it, i had defined  `BABYSTEP_INVERT_Z` as '1' or '-1' and made a direct multiplication, saving the decision. ;-)

Wackerbarth:
But that "in all the configurations" is where the problem exists.
If the entry applies to the default, it will migrate from the multiple configuration files and be placed in the configuration defaults.  Then, what happens when someone wants the inversion to be true?

So, we will have to condition the default on DISABLED(BABYSTEP_INVERT_Z)  (or equivalent).

But, in those cases where we want to specify inversion, isn't it easier to follow the logic and define the proper value if we use ENABLED(BABYSTEP_INVERT_Z) in the code rather than having to remember that BABYSTEP_INVERT_Z is a true/false value rather than just an option?

AnHardt:
As saied before. I did not invent or concepted this. I just made use of a already existing define.

As every other defined by default option it has to be encapsulated in a `#ifndef OPTION\n   #define OPTION value\n#endif` to make it setable by a preloaded customization-file or compiler option. I'm quite sure, i talked about that when you came up with this idea the first time.

In the configs we used to have, we have a context, a example value and ideally a comment. That's not a that bad hint about what the new value should look like.

How, do you suppose the customization-files are written? My best guess is to have a existing configuration file in one window and the new one in an other. Reading in the one and writing to the other.

Wackerbarth:
"As said before. I did not invent or conceive this. I just made use of a already existing define."

If someone made a poor choice in the past, that is no reason to perpetuate it.

Having the default for any option be "false" (preferably signaled by the absence of any related entry) and, when activated, "true" (without needing to specify a value) is less likely to cause users to make an incorrectly formatted entry.  The `ENABLED` macro that Scott developed is quite useful in this process.

I recognize that there are some parameters which cannot be handled that way because they require numerical values (for example, PID parameters). Those we will have to handle with `#ifdef` constructs.

- - -

"How, do you suppose the customization-files are written? My best guess is to have a existing configuration file in one window and the new one in an other. Reading in the one and writing to the other."

The main difference is that I propose having only one copy of the complete reference file. And I propose that that reference copy SHOULD have much better commenting than the limited amount that we currently place in the difficult to maintain multitude of copies (some of which are, invariably, out of date).  Additionally, since the parts of the configuration entries which actually needed is a small subset of the total, it is much easier to find the "active" components because they are not intermixed with "inactive" ones.

AnHardt:
This is not needed when `Babystep_Multiplicator` is set negative.

PR #273, opened 2015-11-07
Rename numbered travis tests to something meaningful

AnHardt:
Rename Travis Tests to more descriptiv names.


Wackerbarth:
This will require a corresponding change to the boards.txt in the support platform.
I'll work on that part of it.

Wackerbarth:
@AnHardt -- Now that the Arduino IDE 1.6.6 has been released, I think we should move to using it for our Travis tests and drop support for all prior versions of the IDE. Do you agree?

AnHardt:
Thanks.
Yes. The actual release of Arduino (1.6.6) is the easiest one to access for the 'users'. We should test with this.

AnHardt:
Only the actual `release` and `1.0.x` branch work better with ARDUINO 1.0.6 under certainly circumstances (dual extruder).

AnHardt:
Update done.

Wackerbarth:
@AnHardt  -- "Only the actual release and 1.0.x branches work better with ARDUINO 1.0.6 under certain circumstances (dual extruder)."

I'm going to open a separate issue to discuss this.

Wackerbarth:
See PR #276

PR #281, opened 2015-11-19
Fixed missing #if in MiniKossel Customization.h file

scas-mdd:
Missing #if ENABLED(PIDTEMP) in MiniKossel customizatio.h file 

https://github.com/MarlinFirmware/MarlinDev/blob/master/configurations/vendor/generic/MiniKossel/Customization.h#L25

I do not have this kind of board, but from my understanding that code block either should be under PIDTEMP condition, or the #endif should be removed. I chose the first one, but I can update the PR if the other one is needed.


scas-mdd:
Ok, did that! I was not aware exactly how it should be (though I looked through other configuration files). Thanks!

PR #282, opened 2015-11-19
[Updated] TotalPrintCount 

iashwinprasada:
as discussed in the main repo's pull request #249 & #237 , I've made some major changes from the latest @RicardoGA  code....

M504 is now kept as a generic code which is called at the beginning and end of each gcoode file.
making it generic opens the door for many such features which can be included at the Start & End easily.

## M504 S E  
+ S - start of gcode
+ E - end of gcode

In this case, we maintain two EEPROM variables to store total prints started and Succecfully ended...
functions to increment it will be inside the code M504 S and M504 E respectively...

Now we can also get the total failed prints
## M505 S C
+ S - get total number of prints and total number of failed prints
+ C - Clear the counters

**tested working on our printer


iashwinprasada:
http://thijs.elenbaas.net/2012/07/extended-eeprom-library-for-arduino/

This still uses struct wise addressing, but additional space for key! How to approach this issue now??

int address = 0;
 
struct keyValuePair {
int key;
float value;
};
 
keyValuePair keyValueInput[2];
keyValuePair keyValueOutput[2];
 
void setup()
{
Serial.begin(9600);
 
keyValueInput[0].key = 2; keyValueInput[0].value = 4.4;
keyValueInput[1].key = 4; keyValueInput[1].value = 8.8;
 
EEPROM.writeBlock(address, keyValueInput,2);
EEPROM.readBlock(address, keyValueOutput,2);

Wackerbarth:
@iashwinprasada -- Your above comment is moving in the right direction.
However, it is still too specific in the sense that it specifies a particular implementation..
We really need to  create some wrapper functions like
```
int get_persistent_integer(LIFETIME_PRINTS_STARTED);
...
void set_persistent_integer(LIFETIME_PRINTS_COMPLETED);
...
``` 
Then you can implement the M- commands using those functions and we can refine the definitions later.

Also, we need to begin working toward splitting up marlin_main.cpp.
Toward that goal, you should place your M-commands in a separate file (related commands in the same file).

Also, with respect to this PR, it needs to be rebased on the current "master" or "dev" so that the diff doesn't include a bunch of unrelated changes that are not a part of your work.

PR #286, opened 2015-11-23
Kc is also a multi-PID parameter

Wackerbarth:
This is the same as Marlin #2794.


Wackerbarth:
#2794 was merged into 1.1. Merging it here to avoid regressing.

PR #291, opened 2015-11-27
Abort on endstop

Wackerbarth:
This is the code that Andreas  proposed to improve end stop handling


AnHardt:
@Wackerbarth 
Thank you for the example integration. Have not analysed this jet. But my code seems to be in. ;-)

Wackerbarth:
/Users/wacky/Documents/Arduino/libraries/MarlinDev/stepper.cpp: In function 'void checkHitEndstops()':
/Users/wacky/Documents/Arduino/libraries/MarlinDev/stepper.cpp:295:30: warning: array subscript is above array bounds [-Warray-bounds]
         axis_known_position[3] = { false }; // not homed anymore

AnHardt:
Yep. Has to be done in a loop. Just used the initialization syntax.

`for (int i = 0; i<3; i++)`axis_known_position[i] = false;

Wackerbarth:
@AnHardt -- This PR is an example of the motivation for placing pending PR threads in the MarlinDev repository rather than in my (or some other user's) personal fork of the repository. By having it in an accessible branch, you were able to make a correction without the hassle of asking me to merge your PR, etc.

Now, the change can be tested and if it works as expected, merged into the main body of code.

a4jp-com:
Very cool ^^

MagoKimbra:
Hi, I checked, but I have a probe optical for autobedlevel. In this way before it arrives at the plate stop printing because it is activated. I modified this way:

    #if ENABLED(ABORT_ON_ENDSTOP_HIT_FEATURE_ENABLED)
      if (abort_on_endstop_hit && !(endstop_hit_bits & BIT(Z_PROBE))) {
        #if ENABLED(SDSUPPORT)
          card.sdprinting = false;
          card.closefile();
        #endif
        axis_known_position[3] = { false }; // not homed anymore
        quickStop(); // kill the planner buffer
        Stop();      // restart by M999
      }
    #endif
   
   endstops_hit_on_purpose();

Wackerbarth:
Rather than checking for, and excluding the probe as an endstop here, wouldn't it make more sense to simply ignore setting the probe bit unless we are actively performing a probe?

AnHardt:
@MagoKimbra 
That's one of the ways you can handle the problem. My suggestion was to handle the probe only when deployed/(explicitly turned on)

MagoKimbra:
It would be better not to set the bit of Z-probe, but no longer read the info. I do not know if they actually serve.

#if ENABLED(Z_MIN_PROBE_ENDSTOP)
      if (endstop_hit_bits & BIT(Z_MIN_PROBE)) {
        SERIAL_ECHOPAIR(" Z_MIN_PROBE=", (float)endstops_trigsteps[Z_AXIS] / axis_steps_per_unit[Z_AXIS]);
        LCD_MESSAGEPGM(MSG_ENDSTOPS_HIT "ZP");
      }
    #endif

thinkyhead:
If the equivalent for this is already merged into RCBugFix then we can close this. That code will get migrated here after the release.

boelle:

jbrazio:

PR #297, opened 2015-12-04
Revert some floats to long for speed

Wackerbarth:
These are improvements proposed by @Wurstnase as improvements to 1.1.0-RC

I think that there may also be additional places where the use of float is not necessary, or really appropriate.


Wackerbarth:
Thanks, @Wurstnase

Wurstnase:
You're welcome :)

PR #318, opened 2015-12-29
Add flag to expose PWM_MOTOR_CURRENT to Configuration_adv.h

tamarintech:
When using a board that supports PWM motor current contorls (ex: MiniRAMBo) users need to edit pins_MINIRAMBO.h to set the motor current.

This setting should be in Configuration_adv.h with the DIGIPOT motor current controls.

Added preprocessor check for flag and, if not present, fallback to defaults define din pins_MINIRAMBO.h.

### DIFF
--- a/configurations/transitional_default_configurations/default/Configuration_adv.h
+++ b/configurations/transitional_default_configurations/default/Configuration_adv.h
@@ -268,6 +268,9 @@
 // Motor Current setting (Only functional when motor driver current ref pins are connected to a digital trimpot on supported boards)
 #define DIGIPOT_MOTOR_CURRENT {135,135,135,135,135} // Values 0-255 (RAMBO 135 = ~0.75A, 185 = ~1A)
 
+// Motor Current controlled via PWM (Only functional when motor driver current is driven by PWM on supported boards)
+#define PWM_MOTOR_CURRENT {1300, 1300, 1250} // Values in milliamps
+
 // uncomment to enable an I2C based DIGIPOT like on the Azteeg X3 Pro
 //#define DIGIPOT_I2C
 // Number of channels available for I2C digipot, For Azteeg X3 Pro we have 8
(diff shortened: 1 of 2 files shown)


tamarintech:
I also added the flag into configurations/transitional_default_configurations/default/Configuration_adv.h

Is that correct?  Any other feedback for this PR?

tamarintech:
Today will be 70 days.
Is there anything else I can do to help?

tamarintech:
I saw the RAMPS 1.3 -> 1.4 changes from thinkyhead merged. :+1: 

Can you help me get this merged, @thinkyhead ?  It seems pretty important to me considering all of the companies using the Mini RAMBo to better support the Mini RAMBo.

thinkyhead:
The MiniRAMBo is already covered. In `pins_MINIRAMBO.h` you will find the line:
```cpp
#define DEFAULT_PWM_MOTOR_CURRENT  {1300, 1300, 1250}
```
and it was already being applied by the old code:
```cpp
int motor_current_setting[3] = DEFAULT_PWM_MOTOR_CURRENT;
```
Anyway, let's talk about this change! I propose to disable this option by default, because its purpose is to override the `DEFAULT_PWM_MOTOR_CURRENT` already set in the board's pins file, and for future boards we might not want to override them with `{1300, 1300, 1250}` in every configuration.

I'll have an updated PR for `RCBugFix` as Exhibit-A shortly…

tamarintech:
Note the line directly above this proposed change in Configuration_adv.h.  DIGIPOTS are a feature of the full sized RAMBo.  This work is nearly identical to what was already done for the RAMBo - except, of course, it's for the Mini.

Should neither of these lines be present since they are both available in pins_<BOARD>.h and users should instead be modifying the pins_BOARD rather than the advanced config?

tamarintech:
@thinkyhead Ping.  Looks like we have this change in RCBugFix, but what about the leftover in Dev?  Any more questions RE: PWM control?

thinkyhead:
Dev is proceeding more slowly, so I'm just sticking PRs there for later, rather than let too many untracked changes pile up. Once I have RC4 out I will pay more attention to the PRs accumulating for dev. At that point members will be discussing the next steps for MarlinDev, as some sweeping changes are being proposed for dev (and 1.2.x).

gddeen:
Yeah!

thinkyhead:
This is superseded by changes to `RCBugFix`, and that code will be getting merged over here pretty soon. At that time we'll make sure this feature up to date. So this can be closed.

PR #321, opened 2015-12-30
modified line 2011 to keep symetry with line 1854 in homeaxis()

lrpirlet:
**Background**
I am using mesh leveling. I have a Prusa I3 printer that loves to lose its mechanical setting. So even when saving to EEPROM, I need to measure all the points (21 in my case) every time I want to print...
This is long and tedious...

So I designed a thing that attaches to a servo. when deployed, the servo brings a surface mounted switch between the hot end and the bed. It is activated when the hot end presses the switch on the bed. That means no x nor y offset. 

Wired in parallel to the normal Z switch, I need no special pin.

I want to probe all mesh points (including home X0, Y0) using G29 ( that invokes G28) with probe deployed... I intended to write something, but found @epatel http://www.thingiverse.com/thing:1120142 and will copy what I need.. 

Using latest MarlinDev, I configured the MANUAL_BED_LEVELING, I configured the Servo, I did NOT configure the AUTO_BED_LEVELING_FEATURE  
```cpp
#define MANUAL_BED_LEVELING  // Add display menu option for bed leveling.
#define MESH_BED_LEVELING    // Enable mesh bed leveling.
...
//#define AUTO_BED_LEVELING_FEATURE // Delete the comment to enable (remove // at the start of the line)
#define DEBUG_LEVELING_FEATURE
#define Z_MIN_PROBE_REPEATABILITY_TEST  // If not commented out, Z-Probe Repeatability test will be included if Auto Bed Leveling is Enabled.
...
#define Z_ENDSTOP_SERVO_NR 0
#define SERVO_ENDSTOP_ANGLES {{0,0}, {0,0}, {30,110}} // X,Y,Z Axis Extend and Retract angles
```
**Issue**
When executing G28, the probe does NOT deploy, but at G28 exit, the probe DOES retract...
In homeaxis() line 1854
```cpp
    #if HAS_SERVO_ENDSTOPS
      // Engage Servo endstop if enabled
      if (axis != Z_AXIS && servo_endstop_id[axis] >= 0)
...
```
In homeaxis() line 2011
```cpp
      #if HAS_SERVO_ENDSTOPS
        // Retract Servo endstop if enabled
        if (servo_endstop_id[axis] >= 0) {
...
```
**My solution**
I suggest keeping the same logic and use the same test to allow servo deploy and servo retract.
Could you consider this pull request that does solve my issue?
Thanks in advance.

**apologies**
I could not find any description to propose a good merge request, nor in the readme, nor in the wiki. I know that some formating may be needed... Is there a description of how to submit correctly a pull request?


Wackerbarth:
@lrpirlet -- You are taking the statement out of context.  IMHO, it is only because the sled deployment occurs immediately before, and correspondingly, later, the sled is "docked" that this line is written as it is.
The net effect of a G28 leaves the probe UN-DEPLOYED. It appears that you proposed change affects that behavior.

I suggest that an appropriate way to handle this is to first separate probe deployment from other activities.

Essentially, there are places where we need to "deploy_probe (if not currently deployed)" and other where we need to "stow_probe (if not currently stowed)". Otherwise, the probe activity should be a separate part of the overall state of the system.

Since we are in the process of unifying the handling of the probe (Roxy's effort) and splitting out the probe handling from the Marlin_main.cpp into (a) separate file(s), I think that it is best to postpone consideration of this request and see how it will fit into our future structure.

I do consider your issue to be a valid "need" that we should handle in some reasonable manner.q
And I thank you for bringing it to our attention.

lrpirlet:
@Wackerbarth ... OK, so this is a feature request... Shall I keep it open, or do you want me to close this...
I still have not found any description for someone to propose a good merge request, that would include the actions to perform... Here post submission I am not sure what to do in case of rejection, delay or inclusion in the main ( here Dev) branch, the latest being the less difficult...
Please teach me how to collaborate best.

Wackerbarth:
@lrpirlet -- Yes, keeping it open is acceptable. Alternately you could re-raise it as a feature request issue.
If you do that, it would be helpful if you describe the "need" (desired behavior) before, or in place of providing a solution. That way, as we revisit it, we will have, up front, a statement of the goal. This will make it easier to consider because a reader will not have to repeatedly dig into the issue just to answer the question "What does he want to accomplish?"

thinkyhead:
Hi @lrpirlet – Are you still pondering this change? If not, perhaps we can close it.

boelle:

lrpirlet:
@boelle Thanks, I should have closed it when someone modified the behavior in Marlin...

PR #346, opened 2016-02-07
Remove obsolete HEATER_BED_DUTY_CYCLE_DIVIDER

thinkyhead:
As noted in MarlinFirmware/Marlin#2908


Wackerbarth:
@thinkyhead -- We need to be making PRs against a branch other than "dev" so that "dev" can be the cleaned up version of the accumulated changes.

Doing so will also allow us to get around the problems caused by the way the Github web interface does merges.

I'm going to merge this one to appease those who want everything to go through that web interface. But I will then rebase the result to reflect the way that otherwise it would have been merged.

There are a long list of pending changes that were on the Jan20 branch which need some review. They are an combination of changes introduced by Roxy and others and, if appropriate, should be included.
However, I find some of them questionable and don't want to include them without some additional review.

Would you please look at them. In particular, see the ones which say "NEEDS REVIEW". They are now on the Feb9 branch.

PR #350, opened 2016-02-08
Apply bed leveling matrix when switching extruders

thinkyhead:
- Redo of #124. See also MarlinFirmware/Marlin#2734

As noted in #122 by @jminardi, the auto bed leveling matrix is not taken into account when switching extruders. This patch applies the corrected difference between extruders.


Wackerbarth:
Although this appears to be an appropriate correction, I think that a solution which tracks position in terms of the location of the tool carrier in the coordinate system of the printer frame will provide a cleaner implementation in the long run. In such a system, both the bed position adjustment and the tool offset corrections would be applied to each object location when converting from object coordinates to printer coordinates.

PR #351, opened 2016-02-10
Enable USE_WATCHDOG by default

thinkyhead:
As suggested by @daid and others, make thermal protection as safe as possible by default.

- Also patch up the commentary for `AUTOTEMP`

### DIFF
--- a/configurations/deprecated_configurations/Felix/Configuration_adv.h
+++ b/configurations/deprecated_configurations/Felix/Configuration_adv.h
@@ -353,7 +353,7 @@
 // @section more
 
 // The hardware watchdog should reset the microcontroller disabling all outputs, in case the firmware gets stuck and doesn't do temperature regulation.
-//#define USE_WATCHDOG
+#define USE_WATCHDOG
 
 #if ENABLED(USE_WATCHDOG)
   // If you have a watchdog reboot in an ArduinoMega2560 then the device will hang forever, as a watchdog reset will leave the watchdog on.
--- a/configurations/deprecated_configurations/Felix/Configuration_adv.h
+++ b/configurations/deprecated_configurations/Felix/Configuration_adv.h
@@ -36,16 +36,6 @@
   #define THERMAL_PROTECTION_BED_HYSTERESIS 2 // Degrees Celsius
 #endif
 
-/**
- * Automatic Temperature:
- * The hotend target temperature is calculated by all the buffered lines of gcode.
- * The maximum buffered steps/sec of the extruder motor is called "se".
- * Start autotemp mode with M109 S<mintemp> B<maxtemp> F<factor>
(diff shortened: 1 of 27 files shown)


Wackerbarth:
No option should be enabled by default. If we want the feature to be the default, then, instead, we should have an option to disable it.

daid:
"disable" ifdefs should be avoided IMHO, as they make code harder to follow. You get into the double negative logic problems, taking more brain power to "solve".

Wackerbarth:
@daid -- First, I question the value in using #ifdef constructs to suppress features such as this. If the user wants to turn the feature off, it is probably because it doesn't function properly for them. In that case, the extra memory required is not likely to be an issue.

Additionally, if it bothers you, you can hide the double negative by having it appear only in the Conditionals.h construct.
```
#ifdef DISABLE_FEATURE_X
#else 
#define USE_FEATURE_X
#endif 
```
and in the code

```
#if ENABLED(USE_FEATURE_X)
<code for feature x here>
#endif 
```

thinkyhead:
> No option should be enabled by default.

I agree in principle. But in that case, do you want to reverse the earlier decision to enable the thermal protection features? It was considered [important enough](https://github.com/MarlinFirmware/Marlin/issues/1932), in the interest of fire safety, to enable `THERMAL_PROTECTION_HOTENDS` and `THERMAL_PROTECTION_HOTENDS` by default. Visions of things burning up convinced me those two were a good idea. Meanwhile, `USE_WATCHDOG` (badly named as it is) is an adjunct to `THERMAL_PROTECTION_HOTENDS`.

See MarlinFirmware/Marlin#1932 for the original discussion.

Wackerbarth:
I was not speaking to the FEATURE, but only to the OPTION related to it.

In this case, IMHO, the OPTION should be to DISABLE_WATCHDOG.

and, if you share @daid's concern for "double negatives", then utilize the "USE_WATCHDOG" (or whatever) selector in the code.

thinkyhead:
It really comes down to a matter of taste. We have done a good job of having only positive settings which enable things when they are uncommented. The only option that follows the other pattern is `DISABLE_M503`, and that's really a sub-option. We have been consistently making it so that top-level features are always enabled by uncommenting the feature switch. So then there is the question of disabling everything by default.

Well, if we abandon the inclusion of any Configuration headers, then all options are already disabled by default. But many options, thermal protection and otherwise, do happen to be enabled in the example configurations. If you build from a Makefile or build with No Configuration, then the thermal protection options —along with every other option— will be disabled.

But the point of having thermal protection "enabled by default" was never to revamp the code so that thermal protection is always enabled but must be disabled with a negative option to turn it off. It was only to enable them by default in the Configuration files as a best-practice example, and to be able to say "see, we told you it was important!"

I don't think we should make _any_ option enabled by default so that it must be explicitly disabled using a `DISABLE_THIS_OPTION` switch.

PR #356, opened 2016-02-11
Differentiate Axis Homed from Axis Unknown

thinkyhead:
This implements the approach discussed at MarlinFirmware/Marlin#2661

- Convert `axis_known_position` to a bit-mask
- Add a new flag `axis_was_homed` and use it to decide how to display LCD coordinates:

Axis not homed: `X:---`
Axis homed and known: `X:123.4`
Axis homed but unknown: `?:123.4`

Note that the XY coordinates on graphical displays are still missing the sign when negative. That will need to be patched separately.


MagoKimbra:
There is an error in the control ...
if (axis_known_position & (BIT(X_AXIS)|BIT(Y_AXIS)) != BIT(X_AXIS)|BIT(Y_AXIS)) {

This is correct
if (axis_known_position & (BIT(X_AXIS)|BIT(Y_AXIS)) != (BIT(X_AXIS)|BIT(Y_AXIS))) {

thinkyhead:
> There is an error in the control…

Holy crap, you're right. This is one of those cases where operator precedence is non-intuitive. You would think that binary operators like `|` and `&` would take precedence over logical operators `||` and `&&`, but no!

thinkyhead:
Compare to the current implementation in the RCBugFix branch. One uses bit fields, the other uses simple boolean arrays. I haven't compared size or performance. They should be comparable, and I have no preference over which method is ultimately used.

jbrazio:

PR #358, opened 2016-02-11
Add CORE support to st_get_position_mm

thinkyhead:
This addresses MarlinFirmware/Marlin#2766 (and MarlinFirmware/Marlin#2995) by adding COREXY and COREXZ support to `st_get_position_mm`. Code that uses this function (including homing and probing routines) should now get the correct axis positions.

The function `st_get_position_mm` has been renamed `st_get_axis_position_mm` so it's clear that the values pertain to the cartesian axes, not necessarily to individual stepper positions.

A second commit adds COREXY and COREXZ support to `M114` and changes the report output for cartesians also, so that they all report raw steps and not millimeters. This is a behavioral change to an _extended, non-standard feature_. (See the `M114` documentation at reprap.org.)

A third commit fixes `endstops_trigsteps` so it uses the steps from both CORE steppers in the calculation. With this patched the `checkHitEndstops` function _should_ display the correct positions.

### DIFF
--- a/stepper.cpp
+++ b/stepper.cpp
@@ -1106,9 +1106,8 @@ void st_set_e_position(const long& e) {
 }
 
 long st_get_position(uint8_t axis) {
-  long count_pos;
   CRITICAL_SECTION_START;
-  count_pos = count_position[axis];
+  long count_pos = count_position[axis];
   CRITICAL_SECTION_END;
   return count_pos;
 }
--- a/stepper.cpp
+++ b/stepper.cpp
@@ -286,6 +286,12 @@ void checkHitEndstops() {
   }
 }
 
+#if ENABLED(COREXY)
+  #define CORE_AXIS_2 B_AXIS
+#elif ENABLED(COREXZ)
+  #define CORE_AXIS_2 C_AXIS
+#endif
+
 void enable_endstops(bool check) { check_endstops = check; }
 
 // Check endstops - Called from ISR!
@@ -312,12 +318,6 @@ inline void update_endstops() {
   #define TEST_ENDSTOP(ENDSTOP) (TEST(current_endstop_bits, ENDSTOP) && TEST(old_endstop_bits, ENDSTOP))
 
   #if ENABLED(COREXY) || ENABLED(COREXZ)
-
-    #if ENABLED(COREXY)
-      #define CORE_AXIS_2 B_AXIS
-    #else
-      #define CORE_AXIS_2 C_AXIS
-    #endif
 
     #define _SET_TRIGSTEPS(AXIS) do { \
         float axis_pos = count_position[_AXIS(AXIS)]; \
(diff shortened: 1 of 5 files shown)


stepper.cpp
```diff
@@ -1087,7 +1105,21 @@ long st_get_position(uint8_t axis) {
...
+    else if (axis == Y_AXIS)
+      axis_pos = (st_get_position(A_AXIS) - axis_pos) / 2; // ((x+y)-(x-y))/2 -> (x+y-x+y)/2 -> (y+y)/2
+  #elif ENABLED(COREXZ)
+    if (axis == X_AXIS)
+      axis_pos = (axis_pos + st_get_position(C_AXIS)) / 2; // ((x+z)+(x-z))/2 -> (x+z+x-z)/2 -> (x+x)/2
+    else if (axis == Z_AXIS)
+      axis_pos = (st_get_position(A_AXIS) - axis_pos) / 2; // ((x+z)-(x-z))/2 -> (x+z-x+z)/2 -> (z+z)/2
+  #endif
+  return axis_pos / axis_steps_per_unit[axis];
+}
```
AnHardt: st_get_position() returns actually `count_position[axis]` in protected manor, so the stepper interrupt can not change them during the read. When calculating with more axes at the same time, it needs to read the values at the same time - if moving.

thinkyhead: Good point! This should be much better then…
```cpp
#if ENABLED(COREXY) || ENABLED(COREXZ)
  #if ENABLED(COREXY)
    #define CORE_AXIS_2 B_AXIS
  #else
    #define CORE_AXIS_2 C_AXIS
  #endif
#endif

void st_get_two_positions(long pos[2], uint8_t axis1, uint8_t axis2) {
  CRITICAL_SECTION_START;
  pos[0] = count_position[axis1];
  pos[1] = count_position[axis2];
  CRITICAL_SECTION_END;
}

float st_get_axis_position_mm(AxisEnum axis) {
  float axis_pos;
  #if ENABLED(COREXY) | ENABLED(COREXZ)
    if (axis == X_AXIS || axis == CORE_AXIS_2) {
      long pos[2];
      st_get_two_positions(pos, A_AXIS, CORE_AXIS_2);
      // ((a1+a2)+(a1-a2))/2 -> (a1+a2+a1-a2)/2 -> (a1+a1)/2 -> a1
      // ((a1+a2)-(a1-a2))/2 -> (a1+a2-a1+a2)/2 -> (a2+a2)/2 -> a2
      axis_pos = (pos[0] + ((axis == X_AXIS) ? pos[1] : -pos[1])) / 2.0f;
    }
    else
      axis_pos = st_get_position(axis);
  #else
    axis_pos = st_get_position(axis);
  #endif
  return axis_pos / axis_steps_per_unit[axis];
}
```

stepper.cpp
```diff
@@ -311,23 +319,38 @@ inline void update_endstops() {
...
-    // Head direction in -X axis for CoreXY bots.
-    // If DeltaX == -DeltaY, the movement is only in Y axis
-    if ((current_block->steps[A_AXIS] != current_block->steps[B_AXIS]) || (TEST(out_bits, A_AXIS) == TEST(out_bits, B_AXIS))) {
-      if (TEST(out_bits, X_HEAD))
-  #elif ENABLED(COREXZ)
-    // Head direction in -X axis for CoreXZ bots.
-    // If DeltaX == -DeltaZ, the movement is only in Z axis
-    if ((current_block->steps[A_AXIS] != current_block->steps[C_AXIS]) || (TEST(out_bits, A_AXIS) == TEST(out_bits, C_AXIS))) {
+    #define _SET_TRIGSTEPS(AXIS) do { \
+        CRITICAL_SECTION_START; \
```
Wurstnase: Is this right? This routine is in an ISR.

thinkyhead: Good catch, thanks!

Wurstnase: You're welcome.

PR #361, opened 2016-02-15
Mixing Extruder Support

thinkyhead:
I was recently contacted by email on the subject of "mixing extruders" and decided to take a look at the requirements for this feature. Based on a proposal by [Pia Taubert](https://www.youtube.com/channel/UCuCd30kb93WIzKjD4W8OECg) in the attached documents at http://reprap.org/wiki/Mixer_extruder, this PR implements the required components to support an extruder with multiple filament inputs and multiple coordinated stepper motors.

- Added new configuration options `MIXING_EXTRUDER_FEATURE` and `MIXING_STEPPERS` to enable a mixing extruder with 2 to 6 stepper motors — theoretically.
- Added `MIXING_VIRTUAL_TOOLS` setting, which enables `M164` command, for compatibility with Repetier Host virtual tools. With this option color mixing is changed by switching tools with the T command.

**Extended GCode**
- Support is added for the [Repetier Host compatible GCode](http://reprap.org/wiki/Repetier_Color_Mixing) `M163 Sn Pn` which sets a single mix factor at a time. For example `M163 S0 P0.5` and `M163 S1 P0.5` would be used to set a mix of half color 1 plus half color 2.
- A shorthand gcode `M165` is proposed to set the mixing parameters globally using parameters `ABCDHI` as a shortcut over multiple `M163`. For example, `M165 A0.5 B0.5` would be used to set a mix of half color 1 plus half color 2.
- If `MIXING_VIRTUAL_TOOLS` is set to 2 or greater the Repetier Host compatible `M164 Sn` command is also made available to save the current mix factors as a virtual tool that can be recalled later. This option changes the behavior of the `gcode_T()` function so it restores a saved mixture rather than setting a new extruder.
- Following Pia Taubert's proposal, the `G1` command is extended to accept up to 6 mixing parameters (`ABCDHI`). For example `G1 A0.25 B0.6 C0.15 X10.0 Y99.2 E12.34 F9000` will set a mixture for 3 channels starting with the current move. The mix parameters must add up to 1.0. If they don't they will be normalized – scaled up or down to add up to 1.0. The mix is persistent, so further moves will continue to use the same mix.

**Code Changes**
- In the planner, the `block_t` structure gains fields to do stepping for mixing steppers. The `plan_buffer_line` function uses the global mixing to convert a single E movement into up to 6 proportional movements.
- The stepper code and supporting macros take account of multiple steppers. Enable/disable/direction changes apply to all steppers simultaneously. Normal stepping and `ADVANCE` stepping both continue to track E steps (not actually stepping), and also apply Bresenham (actually stepping) to the mixing steppers.

**Needed**
- A way to enable only the steppers that will actually be in use, for times when not all inputs to the mixing extruder are needed. This could be automated by setting a timeout on each one, so they only turn on when they need to start stepping, but turn off after a period of inactivity — 5 minutes, for example. This could be checked when adding blocks to the planner rather than expanding the stepper interrupt.

I hope to get lots of instructive feedback, especially if you are using mixing extruders of any kind. This code has not been tested at all yet. https://github.com/thinkyhead/MarlinDev/tree/new_mixing_extruder


Wackerbarth:
@thinkyhead -- I'm concerned about the way that you are "appropriating" the "ABCD" designators for this purpose.  I had considered that it would be appropriate to use those to identify individual actuators for CORE*, DELTA, etc. kinematics to allow proper trim operations on individual steppers to be albe to correct for misalignments such as might be found in a typical DELTA configuration or a configuration which has two steppers which normally move in sync when making a cartesian move. (But, for example, might be moved independently by a small distance as a part of a "leveling" operation).

I think that it would be a good idea to have a more comprehensive DESIGN of the "language" before we get further stuck with incompatible constructs. I feel fairly certain that we can come up with some solution that will allow both kinds of features to coexist in a way that remains logical for the users of either (or both) features.

thinkyhead:
This is an extension that is actually in use on [Builder 3D printers](http://www.dynamism.com/3d-printers/big-builder.shtml?gclid=CP6p6u7v-soCFQ-oaQodoNwE7A), and others. The specific letters used can be easily changed for anyone's preference, but the tools for producing multi-color prints follow the ABC lettering as a de-facto standard. This is a complement to that.

This feature turns out to be remarkably simple and light, thankfully nothing that need topple the code paradigm of Marlin. Since this is only a proposed implementation, we can now refer to it in discussion of specific alterations, refactoring, etc. This is "EXHIBIT A" capice? This is reference code against the current code-base. Its tight integration alongside other legacy features demonstrates, and localizes, the idealized implementation at the level of ISR, planner, etc.

As for the forseeable problem with `DELTA` conflict. The proposed `M649` code is one simple way around that. The `G1` extension may be turned off when and if an extension to `DELTA` is made that needs to actuate the ABC axes, and `M649 ABC` may be used exclusively to alter the color mix. I recommend it above the `G1` method entirely.

It would be nice to have successive letters, but finding more than 5 is difficult … so … `HIJKL` `STUVW`

thinkyhead:
Added support for Repetier Host's `M163` and `M164` commands, and changed the "Set Mix" command to `M165` for better parity with this numbering scheme. The Repetier-compatible codes are:
```
M163 S<stepper> P<weight> : Set the weight (mix factor) for one stepper
M164 S<virtual-tool-id>   : Save the mix as a Virtual Tool
```
See http://reprap.org/wiki/G-code#M163:_Set_weight_of_mixed_material for a description of these codes as they are used by Repetier.

@repetier

thinkyhead:
The main thing needing scrutiny is the methodology used at the stepper and planner level to convert the mix factors (weights) into movements. The technique I used was to _reduce_ the step values based on the mix and use those as the Bresenham seeds against the same Bresenham ceiling as the rest of the axes. But it may be a more accurate method to use the `block->steps[E_AXIS]` as the adder for all three, but instead scale `block->step_event_count` up and use scaled ceilings. The result should end up being more accurate, without adding any extra overhead.

MagoKimbra:
@thinkyhead i test this on my MarlinKimbra
i think found this bug:
for (uint8_t j = 0; j < MIXING_STEPPERS; j++) {
  if (counter_m[j] > 0) { // general idea... do this 4 times
    counter_m[j] -= current_block->step_event_count;
    En_STEP_WRITE(j, INVERT_E_STEP_PIN);
  }
}

This for me is exactly
for (uint8_t j = 0; j < MIXING_STEPPERS; j++) {
  if (counter_m[j] > 0) { // general idea... do this 4 times
    counter_m[j] -= current_block->mix_event_count[j];
    En_STEP_WRITE(j, INVERT_E_STEP_PIN);
  }
}


Please explain me the function for this. I do not understand..
block->mix_event_count[i] = block->step_event_count / mixing_factor[i];

The mixing_factory is possible equal 0, but division for 0 not possible...
Why is not block->step_event_count * mixing_factor[i];

Sorry for my bad english...

thinkyhead:
I believe the divide-by-zero might be avoided because the divisor is guaranteed to always be zero if the mixing factor (dividend) is zero. "Zero divided by zero" is legal, I believe, and might even evaluate to 1. But, rather than chance it, I will add a check for zero there.

As for the way I'm doing the Bresenham on the mix factors, I remember there was some reasoning for using the same "ceiling" as regular movement. But you may be right. I believe I'm scaling up the ceiling for each factor rather than scaling down the addends. So your fix is probably right. But, I will look more closely at it later, after much sleep!

thinkyhead:
Oh, right. Why am I dividing by the mixing factor? Because they will be one or less. Thus dividing by the factor actually **scales up** the ceiling (step event count). I wanted to scale the ceilings up rather than scale the addends down, because it should give more precision.

thinkyhead:
I guess you already figured out that this was silly:
```cpp
#if ENABLED(MIXING_EXTRUDER_FEATURE)
  for (uint8_t j = 0; j < MIXING_STEPPERS; j++)
    e_steps[current_block->active_extruder] += (advance_whole - old_advance) * current_block->mix_factor[j];
#else
  e_steps[current_block->active_extruder] += advance_whole - old_advance;
#endif
```
…because (since the mix factors must add up to 1.0) it's really exactly the same as this:
```
e_steps[current_block->active_extruder] += advance_whole - old_advance;
```

thinkyhead:
I note that the bug you mention has already been patched for a while. This…
```cpp
counter_m[j] -= current_block->step_event_count;
```
…was already replaced by…
```cpp
counter_m[j] -= current_block->mix_event_count[j];
```
…around 2 weeks ago maybe.

MagoKimbra:
I test width the new solution, bat i have this problem...
If block->mix_event_count[i] is equal to 0 all motors move independently from factory mix.
if i write this
`block->mix_event_count[i] = (mixing_factor[i] < 0.0001) ? 999999999999 : block->step_event_count / mixing_factor[i];`
it works

thinkyhead:
You're right, and your solution will work. Actually this would be best:
```cpp
block->mix_event_count[i] = (mixing_factor[i] < 0.0001) ? 0xFFFFFFFF : block->step_event_count / mixing_factor[i];
```
But instead of doing that —making it so the `mix_event_count` can "never" count down to zero— it seems more "proper" to set it to zero and then check for zero, so that's the patch I just made:
```cpp
counter_e += current_block->steps[E_AXIS];
for (uint8_t j = 0; j < MIXING_STEPPERS; j++) {
  if (current_block->mix_event_count[j] != 0) {
    counter_m[j] += current_block->steps[E_AXIS];
    if (counter_m[j] > 0) En_STEP_WRITE(j, !INVERT_E_STEP_PIN);
  }
}
```

jbrazio:

mkc2000:
@ thinkyhead and MagoKimbra
I am using Marlin 1.1.0-RC7 FW
I can get the G1 mixing to function properly but it has to be placed in every extruder line (currently using 3 extruders but I will try 5 if this is successful)...
G1 A0.2 B0.2  C0.6 E10
I have tried the format:
M163 S0 P5
M163 S1 P1
M163 S2 P0
M164 S0
also
M163 S0 P0.2
M163 S1 P0.2
M163 S2 P0.6
M164 S0
then initiating T0 by:
M104 T0 S240
G92 E0
M109 T0 S240
My extruder moves do nothing...

I am using Repetier-Host V1.6.2 and have set my virtual extruders set but the M163, M164 method is not working.  I have also tried a variety of M165 commands :( nothing
Any suggestions?  Also is there a way to have the G1 method inserted only once at each colour change, rather than every extrusion move??
Also a few more questions because I think they will be quick answers for you ;)
What is the max number of MIXING_STEPPERS? 4? I am using Ramps 1.4 with X3 Stepper Expander
In Repetier-Host can I setup 5 scripts of 16 virtual mixing values?
Thanks!

#define MIXING_EXTRUDER
#if ENABLED(MIXING_EXTRUDER)
  #define MIXING_STEPPERS 3        // Number of steppers in your mixing extruder
  #define MIXING_VIRTUAL_TOOLS 16  // Use the Virtual Tool method with M163 and M164
  #define DIRECT_MIXING_IN_G1    // Allow ABCDHI mix factors in G1 movement commands
#endif

lavato:
@thinkyhead 

Hi!

Thank you for putting an effort to do this, it is really much appreciated!!!

What is the status of this PR now? 
Is it part of  RCBugfix ?

thinkyhead:
> Is it part of RCBugfix ?

Yes, it was merged many moons ago.

orhaan68:
Dear friends, i am quite new at 3D printing. I am using marlin 1.1.9 code. I made a printer with diamond hotend with 3 extruder. I configure with mixing extruder on Marlin. I made a color mixing on Repetier software. When i test Steppers for;
M163 S0 P1
M163 S0 P0
M163 S0 P0
M164 S0

M163 S0 P0
M163 S0 P1
M163 S0 P0
M164 S1

M163 S0 P0
M163 S0 P0
M163 S0 P1
M164 S2

It works fine. When i sent T0 E100 the first extruder move 100 mm. Also it works for T1 and T2 with no problem. But for mixing movement as;

M163 S0 P0.33
M163 S0 P0.33
M163 S0 P0.34
M164 S15

Then i sent T15 E100 all three steppers move 75 mm (they have to move 33, 33, and 34 mm). This problem occure for other mixing ratio commands. Steppers move more than mixing ratio. I can not fix this problem. I reflash marlin more times but result is same. I also try G1 E100 A0.33 B0.33 C0.34 command. But again steppers move 75 mm. How could i fix this problem?

thinkyhead:
@orhaan68 — Upgrade to `bugfix-2.0.x`. https://github.com/MarlinFirmware/Marlin/tree/bugfix-2.0.x

PR #383, opened 2016-03-13
Extended M33

thinkyhead:
Currently `M33` simply echoes back the full path. A host works with this by taking any reply starting with `/` slash as a long filename reply, safely assuming that they're in the same order as requested. While this works fine, even asynchronously, it lacks the flavor of a proper protocol.

Furthermore, the full path is returned on every request. This is redundant information once the host knows the path components. Hosts could obtain longname file listings faster if this extraneous information were left out of the reply.

This set of changes extends `M33` with an additional reply format, adding two new parameters.
- `S` takes an integer which will be echoed back with the reply. This allows software to correlate the request with the reply.
- `P0` returns only the filename. Non-zero returns the full path. Default is to return the full path.
- If `S` or `P` parameters are supplied, the reply format will be “`M33 nn /Long/Path/to/File.ext`”.

Examples:
```gcode
> M33 miscel~1/armchair/armcha~1.gco
< /Miscellaneous/Armchair/Armchair.gcode
< ok

> M33 P0 miscel~1/armchair/armcha~1.gco
< M33 0 Armchair.gcode
< ok

> M33 S23 miscel~1/armchair/armcha~1.gco
< M33 23 /Miscellaneous/Armchair/Armchair.gcode
< ok
```


jbrazio:

PR #384, opened 2016-03-13
Option to reverse the click-wheel direction in menus

thinkyhead:
Also see MarlinFirmware/Marlin#3114. In answer to MarlinFirmware/Marlin#1720. Add an option to reverse the menu navigation direction using the click-wheel.

Currently in Marlin: When navigating menus on the LCD you rotate clockwise to move the menu highlight down and counter-clockwise to move the menu highlight up.

Background: Some users have noted that if the click wheel is located to the right of the screen, the edge of the wheel next to the screen will be moving in the opposite direction from the menu navigation. Previously it has been suggested to simply change the pin numbers so that the wheel is reversed, but this results in problems with editing values. So we needed a way to be able to turn this behavior on and off, or to have the menus interpret the wheel direction in a different way from other code.

Investigation: We quickly uncover that there is a single place in the code where wheel pulses are turned into changes in the encoder value, and functions can pre-set the encoder value to whatever they want before reading changes to it. Menu functions, for example, use the current encoder value to decide what to highlight and when to scroll.

The Solution: 1. Add the ability to reverse the direction of the wheel at the lowest level. 2. Reverse the direction in most instances – whenever `START_MENU` is applied. Set it forward for the editing of values, status screen, and any other non-menu contexts.

In addition, this PR fixes a bug:
The parameters to `lcd_goto_menu` changed, but `menu_edit_type` was not updated. So after editing a value the selected item would always be the top. This restores the correct behavior.


jbrazio:

PR #386, opened 2016-03-13
BABYSTEPPING for COREXY and COREXZ

thinkyhead:
Corresponds to MarlinFirmware/Marlin#3127

Background: The `BABYSTEPPING` feature allows you to use the LCD controller to apply steps and adjust positioning at any time, even during a print in progress. The low-level part of this feature applies directly to steppers ("ABC"). The user interface and configuration are both specified to move the "XYZ" axes. So `BABYSTEPPING` is a "convenience" feature, not a "hack" feature. This PR "breaks" the single-stepper aspect of `BABYSTEPPING` but it does the right thing that you would want on a CoreXY or CoreXZ machine, namely applying babysteps to more than one stepper.

This change is applied in the `_lcd_babystep` function and doesn't touch the babystepping handler in the temperature.cpp `ISR(TIMER0_COMPB_vect)`. For a more universal and transparent application of `BABYSTEPPING` the low-level babystep handler could be changed instead. I've left that code alone because there may be reasons we want to keep "untranslated" babystepping for individual steppers in place for a later feature.

References: MarlinFirmware/Marlin#2935, MarlinFirmware/Marlin#3060


jbrazio:

PR #387, opened 2016-03-14
Updated pt-br translation

hydrocat:
As stated in a previous issue, a new file was made to be used in displays capable of writing latin characters (such as ã á ç ) and the "old" language_pt-br.h has "normal" characters only.


thinkyhead:
Do you mind also making this PR for our `RCBugFix` branch? I would like to have this in the next release candidate so that it can perhaps get more feedback. https://github.com/MarlinFirmware/Marlin/tree/RCBugFix

language_pt-br.h
```diff
@@ -16,55 +16,55 @@
...
 #define MSG_MIN                             LCD_STR_THERMOMETER " Min"
 #define MSG_MAX                             LCD_STR_THERMOMETER " Max"
 #define MSG_FACTOR                          LCD_STR_THERMOMETER " Fact"
-#define MSG_AUTOTEMP                        "Autotemp"
-#define MSG_ON                              "On "
-#define MSG_OFF                             "Off"
+#define MSG_AUTOTEMP                        "Temp. Automatica"
+#define MSG_ON                              "Ligado "
+#define MSG_OFF                             "Desligado"
 #define MSG_PID_P                           "PID-P"
```
Blue-Marlin: The space in "On " is ought to overwrite the second 'f' of "Off". This avoids a complete screen refresh on character based displays.
So:
```
#define MSG_ON                              "Ligado   "
#define MSG_OFF                             "Desligado"
```

hydrocat: so it means that all sentences should have 16 characters ? or this should only be taken into account in messages that are thought to be exchanged frequently, like "on" and "off" ?

Blue-Marlin: @hydrocat 
Forget about it.
Control/Filament right side.
The value is now adjusted to the right automatically.
With the spaces you can avoid the 'jumping' of the start position, but it's not longer needed to avoid a "Onf"

PR #389, opened 2016-03-14
Serial line buffer to prevent queue corruption

thinkyhead:
See MarlinFirmware/Marlin#3012.

This PR adds a single line buffer for the serial input stream.

Background: If any code uses `enqueuecommand` or `enqueuecommands_P` while there's a serial stream (as with an active print job), if there is an incomplete command in the command buffer, the injected command will first overwrite the current buffer slot, then the next buffer slot will get only a partial serial command written into it, causing unpredictable results due to stale data in the ring buffer.

This patch adds an intermediate buffer to cache a single line of serial input. This ensures that only complete commands are ever injected into the queue. So there's no incomplete writing state, and calls to `enqueuecommand` can't corrupt some serial command in-progress.

References: MarlinFirmware/Marlin#2911 MarlinFirmware/Marlin#2951 MarlinFirmware/Marlin#2984


jbrazio:

PR #391, opened 2016-03-15
Wrap macros to prevent bad expansions

thinkyhead:
See MarlinFirmware/Marlin#3140 for details. Basically, macros should always be wrapped in parentheses when appearing in situations like these, and many more:
```cpp
#define MY_MACRO 1 + 2 + 3 // users "shouldn't" do this
a = something - (MY_MACRO)
a = something * (MY_MACRO)
a = something / (MY_MACRO)
a = something % (MY_MACRO)
a = (MY_MACRO) * something
a = (MY_MACRO) / something
a = (MY_MACRO) % something
a = (MY_MACRO) ? true : false;
a = MY_MACRO || something;
```
Since novice users are encouraged to edit configurations, this "best practice" helps to prevent wacky behavior that can be very hard to track down, and for which the cause may be easily missed.


jbrazio:

PR #393, opened 2016-03-16
Add MQ2 Smoke detection feature.

Nutz95:
This feature is compaible with MQ2 gaz sensor and will shut down any heating element if it detects any gaz or smoke.
It can also warn user using LCD buzzer of displaying error message on the LCD or even making the LCD blink when any hazzard is detected.
An error is also logged into the console output to notify possible 3Rd party software.

New compatible settings have been updated into the Scalar XL 3D printer profile.

I've checked that this is building using Scalar XL settings and that enabling/disabling settings won't break the compilation.

This code was is working with Marlin 1.1.0 RC3 and should work properly with this dev branch.

### DIFF
--- a/Gaz_Detector.cpp
+++ b/Gaz_Detector.cpp
@@ -1,0 +1,91 @@
+#ifndef MQ2_DETECTOR_CPP
+#define MQ2_DETECTOR_CPP
+#include "MarlinFirmware.h"
+
+
+#if ENABLED (MQ2_GAZ_DETECTOR)
+#include "Gaz_Detector.h"
+
+const float           LPGCurve[3]  =  {2.3,0.21,-0.47};   //two points are taken from the curve. 
+                                                    //with these two points, a line is formed which is "approximately equivalent"
+                                                    //to the original curve. 
+                                                    //data format:{ x, y, slope}; point1: (lg200, 0.21), point2: (lg10000, -0.59)
+const float           COCurve[3]  =  {2.3,0.72,-0.34};    //two points are taken from the curve. 
+                                                    //with these two points, a line is formed which is "approximately equivalent" 
+                                                    //to the original curve.
+                                                    //data format:{ x, y, slope}; point1: (lg200, 0.72), point2: (lg10000,  0.15) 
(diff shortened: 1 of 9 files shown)


Gaz_Detector.cpp
```diff
@@ -0,0 +1,91 @@
...
+ int MQSmokeDetector::ReadDigitalPin(int mq_pin)
+ {
+      return  digitalRead(mq_pin);   // read the input pin
+ }
+
+float MQSmokeDetector::Read(int mq_pin)
+{
+  return ResistanceCalculation(analogRead(mq_pin)); 
+}
+
```
Blue-Marlin: analogRead() uses busy wait for ~104 micro seconds.
That's the reason for we do not use this.
FILAMENT_SENSOR is a proper example how to do this.
```
int analogRead(uint8_t pin)
{
        uint8_t low, high;

        if (pin >= 14) pin -= 14; // allow for channel or pin numbers

        // set the analog reference (high two bits of ADMUX) and select the
        // channel (low 4 bits).  this also sets ADLAR (left-adjust result)
        // to 0 (the default).
        ADMUX = (analog_reference << 6) | (pin & 0x07);

        // start the conversion
        sbi(ADCSRA, ADSC);

        // ADSC is cleared when the conversion finishes
        while (bit_is_set(ADCSRA, ADSC));  //!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!

        // we have to read ADCL first; doing so locks both ADCL
        // and ADCH until ADCH is read.  reading ADCL second would
        // cause the results of each conversion to be discarded,
        // as ADCL and ADCH would be locked when it completed.
        low  = ADCL;
        high = ADCH;

        // combine the two bytes
        return (high << 8) | low;
}
```

Nutz95:
OK I'm looking at it.
it seems less straight forward.
But i should be able to fit with what is existing.

boelle:

jbrazio:

PR #405, opened 2016-04-19
Updated PID Autotune, added different tuning modes

Pwuts:
There are now 3 tuning modes:
1. Classic PID (autotuning like it was before)
2. Some overshoot: This limits the overshoot during heating and makes it more stable earlier.
3. No overshoot: This almost completely eliminates overshoot during heating and gives (on average) the quickest heat-up time. Disadvantage is that it doesn't allow a heater/thermistor test (check if it will remain stable at high temperatures.)

note: This patch was based on v1.1.0 RC3, I did NOT write line 352-364.


Pwuts:
Forgot to update the other involved files for handling of the M303 command. Reposting corrected patch.

PR #406, opened 2016-04-19
Updated the PID Autotune function with 3 modes

Pwuts:
It now has 3 modes, which are selectable with the M argument:
M1 = Classic PID tuning, just like before
M2 = Some overshoot, limits overshoot and smoothens the stabilization
M3 = No overshoot, eliminates overshoot almost entirely and speeds + smoothens the stabilization.


Marlin_main.cpp
```diff
@@ -4945,11 +4945,12 @@ inline void gcode_M226() {
 inline void gcode_M303() {
   int e = code_seen('E') ? code_value_short() : 0;
   int c = code_seen('C') ? code_value_short() : 5;
+  int mode = code_seen('M') ? code_value_short() : 0;
```
thinkyhead: If this argument is optional then it should be:
```cpp
int mode = code_seen('M') ? code_value_short() : 1; // default to Classic PID
```

temperature.cpp
```diff
@@ -224,6 +224,20 @@ void PID_autotune(float temp, int extruder, int ncycles) {
...
     return;
   }
+  
+  //Tuning mode selection warnings
+  if (mode > 3) {
+    SERIAL_ECHOLN("Bad mode number, only 1, 2 or 3 allowed");
+    return;
+  }
+  
+  if (mode == 0) {
```
thinkyhead: This won't be needed when `M` is optional and defaults to `1`. Then the test above will need to be `if (mode < 1 || mode > 3)`.

temperature.cpp
```diff
@@ -287,30 +301,30 @@ void PID_autotune(float temp, int extruder, int ncycles) {
               Tu = ((float)(t_low + t_high) / 1000.0);
               SERIAL_PROTOCOLPGM(MSG_KU); SERIAL_PROTOCOL(Ku);
               SERIAL_PROTOCOLPGM(MSG_TU); SERIAL_PROTOCOLLN(Tu);
-              Kp = 0.6 * Ku;
-              Ki = 2 * Kp / Tu;
-              Kd = Kp * Tu / 8;
+              //Mode 1, classic PID tuning
+              if(mode == 1){
```
thinkyhead: Two recommendations here:
- Instead of using 1, 2, and 3, make an `enum` and give names to these values. This avoids us needing to know all the "magic numbers" in the code.
- Use a `switch (mode) { ... }` block instead.

Marlin_main.cpp
Pwuts: Nope: if no mode is selected, it replies a request to select a mode 1, 2 or 3, with descriptions of the modes.

Pwuts:
I don't get why it doesn't compile, I didn't change anything to the code in the sections where it gives errors... (but I'm a n00b with C(++) so maybe I'm overlooking something)

Blue-Marlin:
Likely a missed '}'.

Failing with only sending a message is acceptable as long M303 can not be used in a script.
Since we have a 'U' parameter now in RC6/RCBugFix (Marlin_main.cpp ~L5610++) what uses the PID-values immediately, a use in a script is possible now.
When usable in a script it ether needs much more radical ways to fail, or a default value.

Pwuts:
Ok, I'll set the default to 2 because that often gives the nicest results.

Blue-Marlin:
> if(mode == 3){
is not properly closed.

I'd suggest 1 to not change the former behaviour.

Pwuts:
I'd just found it, thanks anyway. :)

mode 1 (the original tuning mode) gives around 50 degrees overshoot on multiple printers, which triggers max overshoot. Furthermore, mode 2 reaches the set temperature more quickly and will have it stable earlier, which is why I think mode 2 is more suitable as a default mode.

Blue-Marlin:
@Pwuts 
> max overshoot error
Whow. Only once heard about that. That was a very unlucky heater with a big distance between heater and thermistor and a huge thermal mass.

@all 
Any related opinions, experiences about the PID-tuning out there?

AnHardt:
Ok. I can reproduce that errors when i do mad things with completely overpowered systems.
Trying to tune a 40W nozzle heater for 50° for example. Or trying to tune the 500W bed heater with `MAX_BED_POWER 255`.

I made some interesting experiments with a M303 with M and U parameter (https://github.com/AnHardt/Marlin/pull/33)
Test-g-code: [pidtest.g.txt](https://github.com/MarlinFirmware/MarlinDev/files/252144/pidtest.g.txt)

Got some surprising results. In the graphs the order is always: tune with M1, use M1, tune with M2 use M2, tune with M3, use M3.

PID_FUNCTIONAL_RANGE 10
pid_functional_range 10

PID_FUNCTIONAL_RANGE 20
pid_functional_range 20

PID_FUNCTIONAL_RANGE 100
pid_functional_range 100


Conclusion:
* PID_FUNCTIONAL_RANGE has still a impact on the PID-algorithm and is worth tuning it
* The different tuning modes do not always do what they promise. Seems to work best if the PID_FUNCTIONAL_RANGE is about right.

jbrazio:
@AnHardt which software are you using to plot the data ?

AnHardt:
@jbrazio 
Manually cropped RepetiorHost screenshots.

AnHardt:
It could be, that the maximum overshot at the beginning of M303 is about the right amount for 'PID_FUNCTIONAL_RANGE'. Will make more tests.

AnHardt:
PID_FUNCTIONAL_RANGE 5
pid_functional_range 5
That's not enough range. PID kicks in to late. Much overshot. Again M1 best, M3 worst.

jbrazio:
If I understood a discussion with @Blue-Marlin about PID, if the delta between current temp and target temp is greater or equal to `PID_FUNCTIONAL_RANGE` then the heating element on with fully PWM thus disabling the PID, the the delta enters inside the range then PID will take over and start PWM de heater.. right @AnHardt ?

temperature.cpp
```diff
@@ -287,30 +301,31 @@ void PID_autotune(float temp, int extruder, int ncycles) {
...
               SERIAL_PROTOCOLLNPGM(MSG_CLASSIC_PID);
-              SERIAL_PROTOCOLPGM(MSG_KP); SERIAL_PROTOCOLLN(Kp);
-              SERIAL_PROTOCOLPGM(MSG_KI); SERIAL_PROTOCOLLN(Ki);
-              SERIAL_PROTOCOLPGM(MSG_KD); SERIAL_PROTOCOLLN(Kd);
-              /*
+              }
+              //Mode 2, some overshoot
+              if(mode == 2){
               Kp = 0.33*Ku;
               Ki = Kp/Tu;
```
AnHardt: According to the literature `Ki = Kp/Tu;` for 'some overshoot'  should also be `Ki = 2*Kp/Tu;`. But changing this gives no better results.

boelle:

jbrazio:

PR #409, opened 2016-04-24
Update MarlinDev with all updates from RCBugFix

thinkyhead:
- All in one large commit, for testing purposes only. Don't merge!


Sniffle:
You poor thing... Good luck

thinkyhead:
I think it will be easier to just rearrange `RCBugFix` into folders in the style of `MarlinKimbra` and then piecemeal bring in the fancy new things loitering in `dev` such as that new experimental configuration system.

jbrazio:
I agree @thinkyhead.

thinkyhead:
@jbrazio Yes, and that way we also can preserve the full commit history (maybe even PR merge commits) as it is.

Issue #379, opened 2016-03-10
State machine and events


jbrazio:
--- Copy/paste from MarlinFirmware/Marlin#3118 ---

A 3D printer is in it's core a state machine:
- Or is idle,
- Or is in the planner doing a move,
- Or is heating the nozzle,
- Or is doing two, three, four things at the same time but all of them are finite.

Currently we have no exact way to know at a specific point in time what is the state of our printer unless we do a over complicated spaghetti check. I found PR #3109 from @thinkyhead where he introduces states brilliant but it has one key aspect missing, our state machine can be doing more than on thing at a time. I would propose the introduction of a global (32 bit ?) "register", we can then specify a bit per state like so [this is not by any means a final list, just food for thought]:

```
#define READY           0x01        // 0000 0000 0000 0000 0000 0000 0000 0001
#define BUSY            0x02        // 0000 0000 0000 0000 0000 0000 0000 0010
#define FAULT           0x04        // 0000 0000 0000 0000 0000 0000 0000 0100
#define HOMED           0x08        // 0000 0000 0000 0000 0000 0000 0000 1000
#define PRINTING        0x20        // 0000 0000 0000 0000 0000 0000 0001 0000
#define WAITING         0x20        // 0000 0000 0000 0000 0000 0000 0010 0000
#define WAITING_USER    0x40        // 0000 0000 0000 0000 0000 0000 0100 0000
#define IN_PLANNER      0x80        // 0000 0000 0000 0000 0000 0000 1000 0000

#define SDCARD          0x100       // 0000 0000 0000 0000 0000 0001 0000 0000
#define SDCARD_BUSY     0x200       // 0000 0000 0000 0000 0000 0010 0000 0000

#define EXTRUDING_0     0x400       // 0000 0000 0000 0000 0000 0100 0000 0000
#define EXTRUDING_1     0x800       // 0000 0000 0000 0000 0000 1000 0000 0000
#define EXTRUDING_2     0x1000      // 0000 0000 0000 0000 0001 0000 0000 0000
#define EXTRUDING_3     0x2000      // 0000 0000 0000 0000 0010 0000 0000 0000

#define HEATING_0       0x4000      // 0000 0000 0000 0000 0100 0000 0000 0000
#define HEATING_1       0x8000      // 0000 0000 0000 0000 1000 0000 0000 0000
#define HEATING_2       0x10000     // 0000 0000 0000 0001 0000 0000 0000 0000
[...]

PR #416, opened 2016-05-02
Introduce M108 cancel wait for heatup and cooldown of the hotend and bed

scalez:
This G-code is asynchronously handled in the `get_serial_commands()` parser.

Continued discussion from https://github.com/MarlinFirmware/Marlin/pull/3611.

### DIFF
--- a/Marlin_main.cpp
+++ b/Marlin_main.cpp
@@ -139,6 +139,7 @@
  * M105 - Read current temp
  * M106 - Fan on
  * M107 - Fan off
+ * M108 - Cancel heatup and wait for the hotend and bed, this G-code is asynchronously handled in the get_serial_commands() parser
  * M109 - Sxxx Wait for extruder current temp to reach target temp. Waits only when heating
  *        Rxxx Wait for extruder current temp to reach target temp. Waits when heating and cooling
  *        IF AUTOTEMP is enabled, S<mintemp> B<maxtemp> F<factor>. Exit autotemp by any M109 without F
@@ -271,7 +272,7 @@ float max_pos[3] = { X_MAX_POS, Y_MAX_POS, Z_MAX_POS };
 
 uint8_t active_extruder = 0;
 int fanSpeed = 0;
-bool cancel_heatup = false;
+bool cancel_wait_heatup = false;
 
 const char errormagic[] PROGMEM = "Error:";
 const char echomagic[] PROGMEM = "echo:";
@@ -897,6 +898,7 @@ void get_command() {
 
       // If command was e-stop process now
       if (strcmp(command, "M112") == 0) kill(PSTR(MSG_KILLED));
+      if (strcmp(command, "M108") == 0) cancel_wait_heatup = true;
 
       cmd_queue_index_w = (cmd_queue_index_w + 1) % BUFSIZE;
       commands_in_queue += 1;
@@ -3914,6 +3916,13 @@ inline void gcode_M105() {
 #endif // HAS_FAN
 
 /**
+ * M108: Cancel heatup and wait for the hotend and bed, this G-code is asynchronously handled in the get_serial_commands() parser
+ */
+inline void gcode_M108() {
+  cancel_wait_heatup = true;
+}
+
+/**
  * M109: Sxxx Wait for extruder(s) to reach temperature. Waits only when heating.
  *       Rxxx Wait for extruder(s) to reach temperature. Waits when heating and cooling.
  */
@@ -3954,9 +3963,9 @@ inline void gcode_M109() {
     #define TEMP_CONDITIONS (degHotend(target_extruder) != degTargetHotend(target_extruder))
   #endif //TEMP_RESIDENCY_TIME
 
-  cancel_heatup = false;
+  cancel_wait_heatup = false;
   millis_t now = millis(), next_temp_ms = now + 1000UL;
-  while (!cancel_heatup && TEMP_CONDITIONS) {
+  while (!cancel_wait_heatup && TEMP_CONDITIONS) {
     now = millis();
     if (now > next_temp_ms) { //Print temp & remaining time every 1s while waiting
       next_temp_ms = now + 1000UL;
@@ -3986,7 +3995,7 @@ inline void gcode_M109() {
         residency_start_ms = millis();
     #endif //TEMP_RESIDENCY_TIME
 
-  } // while(!cancel_heatup && TEMP_CONDITIONS)
+  } // while(!cancel_wait_heatup && TEMP_CONDITIONS)
 
   LCD_MESSAGEPGM(MSG_HEATING_COMPLETE);
   refresh_cmd_timeout();
@@ -4010,9 +4019,9 @@ inline void gcode_M109() {
     // Exit if the temperature is above target and not waiting for cooling
     if (no_wait_for_cooling && !isHeatingBed()) return;
 
-    cancel_heatup = false;
+    cancel_wait_heatup = false;
     millis_t now = millis(), next_temp_ms = now + 1000UL;
-    while (!cancel_heatup && degTargetBed() != degBed()) {
+    while (!cancel_wait_heatup && degTargetBed() != degBed()) {
       millis_t now = millis();
       if (now > next_temp_ms) { //Print Temp Reading every 1 second while heating up.
         next_temp_ms = now + 1000UL;
@@ -5920,6 +5929,10 @@ void process_next_command() {
       case 105: // M105: Read current temperature
         gcode_M105();
         return; // "ok" already printed
+
+      case 108:
+        gcode_M108();
+        break;
 
       case 109: // M109: Wait for temperature
         gcode_M109();
--- a/Marlin_main.cpp
+++ b/Marlin_main.cpp
@@ -139,7 +139,7 @@
  * M105 - Read current temp
  * M106 - Fan on
  * M107 - Fan off
- * M108 - Cancel heatup and wait for the hotend and bed, this G-code is asynchronously handled in the get_serial_commands() parser
+ * M108 - Cancel wait for heatup and cooldown of the hotend and bed, this G-code is asynchronously handled in the get_serial_commands() parser
  * M109 - Sxxx Wait for extruder current temp to reach target temp. Waits only when heating
  *        Rxxx Wait for extruder current temp to reach target temp. Waits when heating and cooling
  *        IF AUTOTEMP is enabled, S<mintemp> B<maxtemp> F<factor>. Exit autotemp by any M109 without F
@@ -3916,7 +3916,7 @@ inline void gcode_M105() {
 #endif // HAS_FAN
 
 /**
- * M108: Cancel heatup and wait for the hotend and bed, this G-code is asynchronously handled in the get_serial_commands() parser
+ * M108: Cancel wait for heatup and cooldown of the hotend and bed, this G-code is asynchronously handled in the get_serial_commands() parser
  */
 inline void gcode_M108() {
   cancel_wait_heatup = true;
--- a/ultralcd.cpp
+++ b/ultralcd.cpp
@@ -390,7 +390,7 @@ static void lcd_return_to_status() { lcd_goto_menu(lcd_status_screen); }
     card.sdprinting = false;
     card.closefile();
     autotempShutdown();
-    cancel_heatup = true;
+    cancel_wait_heatup = true;
     lcd_setstatus(MSG_PRINT_ABORTED, true);
   }
 
--- a/ultralcd.h
+++ b/ultralcd.h
@@ -52,7 +52,7 @@
   extern int absPreheatHPBTemp;
   extern int absPreheatFanSpeed;
 
-  extern bool cancel_heatup;
+  extern bool cancel_wait_heatup;
 
   #if ENABLED(FILAMENT_LCD_DISPLAY)
     extern millis_t previous_lcd_status_ms;


scalez:
Outside of `cancel_wait_heatup = true`, in order to successfully stop a print `DISABLE_STEPPER_DRIVER_INTERRUPT()` needs to be called and the block buffer needs needs to be discarded as is done by `quickStop()` in stepper.cpp.

Essentially, to effectively stop a print through a host you would need to be able to do exactly what `lcd_sdcard_stop()` does inside [ultralcd.cpp L338](https://github.com/MarlinFirmware/MarlinDev/blob/dev/ultralcd.cpp#L388), and you would need to be able to do it asynchronously.

kakaroto:
Answering here since the PR has been moved.

@repetier @foosel @daid : thanks for your comments and opinions on this issue (https://github.com/MarlinFirmware/Marlin/pull/3611). We appreciate your involvement.

First, I apologize for the long comment below, hopefully you have the courage to read it all (if not, at least skip the my response to @Blue-marlin below) :)

I'll now answer everyone's concerns : 
@repetier The implementation we did in the host should work regardless of the flavor of firmware on the machine, it doesn't do anything special like what you're mentioning, it doesn't keep track bytes sent or anything like that. It does however keep track of how many commands have been sent without receiving an 'ok' for them yet. The usual ping/pong you mentioned of sending one command after receiving an 'ok' is only useful if you don't want to support any look-ahead for optimizing moves, but if you do, then you have to send more than one command at a time.
There are 2 assumptions that are made in our host implementation with regards to the firmware. The first assumption is that it supports a command queue (sending more than 1 command at a time) and the second assumption is that it can hold at least 3 elements. 
This is the same assumption required for doing look-ahead, so this is nothing new and everyone already does it.
However, in the case of M108 handling with M109, since we don't send any command after M109, then the firmware doesn't need to have a command queue for it to work, as long as it supports async handling of M108. So the only thing we do different for supporting M108 is to disable look-ahead (sending of multiple commands) after M109 is sent.
Your question about the slowdown is that there is no slow down. The look-ahead is still implemented and it's only disabled when M109/M190 enters the queue, in which case, it will have no impact because the firmware will enter a wait state, so whether it takes 5 minutes to reach 240C or 5 minutes and 100 milliseconds, it won't make a difference. As soon as M109/M190 is done, the command queue is filled again by the host, and look-ahead is re-enabled, so there is no slowdowns for the print itself.
As for the order of commands, the order is not maintained only for async commands (M108/M112) which are already interpreted async in the firmware, therefore their order really doesn't matter. As for the line numbers, they are only sent for gcode from the file itself, and there will never be an M108/M112 command in a gcode file since those are interactive commands that will always be sent by the user. The custom commands (not from gcode file) are sent without a line number and therefore, they will not affect the line number order in Marlin and there will be no 'resend'.

@foosel Thank you, power that be, for allowing us to control the firmware and the software! :)
I know what firmware bug you're referring to and I had to deal with that too, it's basically G0 and G1 sending the 'ok' async. It doesn't really affect us that much though and I have fallbacks in place for that scenario, but I see what you mean and I've been there.
It's actually not a 'bug' but a 'feature', because most host software would only send a new command when it got an 'ok', marlin decided to send the 'ok' asynchronously for moves so that hosts would keep sending new move commands until command queue is full, this allowed marlin to do look-ahead when the host didn't support it. Since then, they removed it and now I think most hosts (I at least know that cura does) will send a few commands (usually 4) to Marlin for to the look-ahead.
This is the only assumption we do in Cura, what M108/M109 handling does is to simply disable the look-ahead until M109 is done, which should work with any firmwares anyways.
I do admit though that while it has been rigorously tested with Marlin (broken and the newer one), it hasn't been tested enough on other firmwares.
On a personal note: sorry if I seem to have offended you. I'm not trying to shoot down any messengers. I am also trying to make things work for the users so they don't have to think about gcodes and communication issues. If you felt offended when I said hosts doing something stupid, I was referring to the clogging of the rx_buffer by spamming serial with M105 while the firmware is busy doing something that takes more than the 5 second timeout, which I've seen Pronterface do quite often. It doesn't mean that the developer is stupid, it's just a use case that was never really tested/investigated/discussed before.

@jbrazio 
> Replying to all in a overview resumé, I do agree that hosts cannot [and should not] try to manage firmware queues, hosts must send the commands as fast as the firmware "requests" them, currently this is with the "OK" reply but if anyone has a better idea I believe we have the right people on the table.

I agree, and we're not managing the firmware queues, we simply stop filling it when we send M109/M190 so we can keep a spot for M108, since there is no need for look-ahead while heating. The keeping track of which command is in the queue is actually to avoid another issue which is about timeouts and flooding the rxbuffer (we basically increase the timeout for not receiving an 'ok' when a G28/G29/M109/M190 is being executed because it usually takes more than the default 10 second timeout). 
I also think that ADVANCED_OK is awesome and we will probably use it in the future. I do have a suggestion for better handling of the command queue and command timeouts, but it needs to be fleshed out/discussed then we'll open an issue so it can be discussed publicly.

@Blue-Marlin You bring a very interesting question here and I think we probably should have started off with that : What is M108 good for?
So you mentioned 3 use cases where we want to stop the heating, here's some others that came from testing or user complaints (assume the printer has no LCD interface) :
* You start a print then decide to print something else, you cancel the print (while it was heating) then try to print something else or you close your host and reopen it, you can't connect to your printer anymore because it's busy, it never sends 'start', it never replies to 'M105' (which is used to detect if the correct baudrate is selected when autodetecting baudrate), etc.. only way to connect to the printer is to send a reset or (what I usually do) manually turn its power off and on again. This is not at all an optimal way to use your printer and I don't want to be the one to explain it in the user manual (not all printers have an on/off switch, and not all printers are easily accessible either). It is not guaranteed that establishing a USB connection will cause a reset on the firmware, and you'd have no way to connect to your printer anymore without a manual reset.
* You start a print then decide to cancel it, you press cancel and walk away from the PC expecting it to cooldown and turn off motors, as usual. The Cancel will send the cooldown commands but they are ignored because it's busy in M109. Host can either send all the commands to the firmware now or queue them until it receives 'ok' responses. If it waits until it receives the 'ok', then if the user closes the host after pressing cancel, then they are all lost and the hotend will reach full temperature  and stay that way overnight while the user is sleeping and expecting it to be off since he canceled it (fire hazard)!
If the host sends it right away, the command queue gets full, the rx_buffer gets full, we overflow the rx_buffer and some commands never make it to the printer (and since they are 'custom', not sent from the gcode file, they don't come with a line number for which the firmware can ask a resend). We end up with the same scenario (in our tests, especially in a dual extruder environment, but also very often in a single extruder environment, the cooldown of at least one of the hotends never reaches the firmware, as it gets dropped from a rx_buffer overflow).
The only way to fix it is to asynchronously tell the firmware to stop heating, so it can accept our next commands and send us 'ok' so we can do our full cooldown sequence.
* You start a print then decide to print something else, you cancel the print (while it was heating) then try to print something else, you have to wait until it reaches target temp before the next print starts (which in our case is a problem because start gcode needs to set temp to 150C for nozzle cleaning, so now we have to wait for it to cooldown again)
* You start a print, then immediately realize you selected the wrong material, you cancel, but hotend keeps heating up even though you canceled (which can be dangerous if you start touching it and expect it to be cold but it's not, or it can damage filament/nozzle if it heats up to the wrong temperature anyway)
* heat bed heats up to 110C but due to badly tuned (or impossible to tune) PID, it reaches 100 and becomes extremely slow for it to reach the last few degrees (and you don't have hysteresis enabled). I used to go in the LCD and change the temp down to 90 manually so the heatup stops waiting, then set it back to 110C so it can keep heating up while the print starts (that was before stop heating was added to lcd menu). That's not possible on the lulzbot Mini which doesn't have an LCD, so then you're stuck in that use case. Being able to just send M108 manually would be a great solution instead.
* You manually enter gcode in your console to control your printer, you use 'M109 S240' then realize you wanted to set it to 140C and you mistakenly used M109 instead of M104, you have no way of cancelling that, so you're forced to unplug your printer and plug it back in again, instead of just sending M108 to cancel that. Definitely not an optimal way of doing things.
* You print with wood filament and used a script (cura has a plugin for that) or manually modify the gcode to make it vary temperature per layer so it can have different coloring per layer for some nice wood effect (and for some reason, you want the temp to reach the target before each layer, instead of a set-it-and-forget-it method, so you can have sharp layer colors instead of gradual change in color of the filament), then at some point, you realize, you set the temp to 2400 instead of 240, you can't cancel it, you have to destroy your 10 hours print instead of just being able to send M108 to cancel the waiting and follow it with the proper target temp. (This is a use case I just thought of, not one that anyone complained about/experienced as far as I know)

You gave a few possible solutions to canceling the heatup that I will now answer (assume no LCD and/or remote control of the printer) :
> Wall-plug

Not a good solution from a user-friendliness point...
> Main-switch

Not a good solution from a user-friendliness point...
> Reset button

Not a good solution from a user-friendliness point... also, may be hard to access in some printers (some have the controller under the bed for example)
> Kill button (hard to continue)

Not all printers have that.
> Modified kill button (press once to finishing the current command, twice to kill) (tbd To Be Done)

Not all printers have that.
> Extra menu item. (tbd)

We're assuming printer has no LCD or that we are remotely controlling it using octoprint.
> Reset via serial, (if the hardware permits that)

Doable but resetting the whole thing is not the same as cancelling the heatup. Imagine the use case of wanting it to cancel heating up the bed so it can just start the print even if the bed is not at final temp. the reset would make it lose its homing and ABL probing.

> Sending M104/M140 with S0 (currently not executed before M109/M190 has finished)

That could be a good solution as a replacement to M108, but you still need to make M104/M140 work asynchronously, which is not a good idea (and still doesn't solve this whole discussion about host/fw comms and async commands), and if we have a gcode that sets temp to 150C for nozzle wiping for example, waits, and then sets it to 240C before it does the wipe, would the M104 in the gcode file be interpreted as "wait for 150C to be reached, then set it to 240C" or "cancel the 150C wait and set it to 240C instead" ? It just adds too much complexity

> Send M112 (currently has to be supported by the host. Otherwise will not work when the buffers are full)

As stated before, you may not want to kill the print, just cancel the never-ending/long wait.

> Sending M108. (tbd) (currently has to be supported by the host - like M112)

Yes! I don't see the problem with this one! :D
I actually don't see any issues with this one apart from the fact that some hosts will have to change the way they handle things in order to support it... but without it, they still would have no acceptable alternative.

> Sending ctrl-c, Tested before put into the RX-buffer - in the RX interrupt. (tbd) (problems with AT90USB processors (USBCON) have to be solved) (has to be supported by the hosts)

I don't know about this, so if you think it's a viable solution, and doesn't contradict any of the above mentioned rebuttals, then I'd like to hear more about it.

> My conclusion:
> Don't merge until it's clear what it is good for. Currently not working with almost all hosts, is not a killer argument, for the integration (See M112), but not helpfull to find reasons for merging. Using a new g-code is always a thing we should think about twice.

Hopefully, I answered what it's good for. Also, I still (sorry) don't see why the comm issues would be a blocker to merging this feature. And the comm issues we mentioned here are causing a lot more problems than just preventing M112 from working (such as the loss of commands as explained above, or M105 getting cropped in the rx_buffer and it becomes M1 which causes a "waiting for user input" in the LCD... VERY easy to reproduce with pronterface).

@daid: dynamically changing M109/M190 into M104/M140 in the host is a good alternative, but it does come with its own drawbacks.. you'd have to implement the whole logic in the host, the hysteresis settings of the user would need to be duplicated or fetched from eeprom prior to a M109/M190 (M501 doesn't even return that info), otherwise the user might start complaining that his hysteresis settings are not working properly in some scenarios. The issue with printing from SD as mentioned by @foosel is also an issue... Overall, it's a workaround rather than a solution to the problems mentioned above.

Also, having the host and printing from SD is a very common scenario, not only that, but I don't see any reason for it not to work, and from a user-friendliness point of view, you can't expect to tell your users that they are doing it wrong because of one bug that can be fixed easily.

As for the BFB conflict, BFB is intended for a 3D systems proprietary machine which isn't even in production anymore. Also, the conflict is only in the gcode command, not in the behavior, sending M108 to BFB will have no effect and receiving a M108 in Marlin that was meant for BFB will also have no effect.

repetier:
> As for the line numbers, they are only sent for gcode from the file itself, and there will never be an M108/M112 command in a gcode file since those are interactive commands that will always be sent by the user. The custom commands (not from gcode file) are sent without a line number and therefore, they will not affect the line number order in Marlin and there will be no 'resend'.

That is not completely true. Repetier-Host/Server always adds a line number (except M117 for marlin problems) even for commands send manually. We keep track of number of bytes send and not lines send. This requires a good bookkeeping for bytes that are freed. So we need to match each ok with the line. That is always the oldest line inserted. If we get a line number we free all lines until that number to correct missed "ok". If we would get out of order "ok" responses the system might fail as it would delete the wrong number of bytes. So even with multiple buffer preserving the order of "Ok" is important. Since it only returns "ok" after input buffer of serial is freed you cache even more with this system - you cache your buffers plus serial input buffer.

In what we refer ping pong mode it is more relaxed. Here we send only after a ok, so buffers returning "ok" early would not be a problem. But in this scenario a missed "ok" can not be catched and will block until host gets a "wait" or times out.

Your solution to stop sending on M109/M190 is the same I proposed to allow parsing of "priority" commands, which should then be send independently of a free "ok" as we assume through wait that there is place enough or we know since anything but M109 is acknowledged.

So as I understand the new system breaks order of "ok" under some circumstances and makes it impossible to count correctly since we can not match them to the line we send. That might cause a comm error with triggered resend to fix it. But the number of these commands is low and rarely used, so we could ignore that problem.

thinkyhead:
> marlin decided to send the 'ok' asynchronously for moves

No, the "ok" response is sent according to the same policy as every other command. The `G1` command is considered "processed" as soon as the move has been added to the queue. The "ok" response indicates that the command has been understood, accepted, and the move is now queued in the planner. It does not indicate that the move has been started or finished (try making a firmware that works that way and see how it performs). This is not a Marlin thing. It's the way a real-time robot firmware must work.

thinkyhead:
> you're forced to unplug your printer and plug it back in again

Actually if the host disconnects and reconnects it re-initializes the printer. So there's no need to physically unplug or hit the power switch.

thinkyhead:
> the new system breaks order of "ok" under some circumstances

@repetier The way `M112` (and the proposed `M108`) currently work is that —yes— the command is processed early, before entering the queue. But the command is still queued immediately afterward, then executed once again when the queue is processed. The "ok" comes only after the queue-processed command returns.

scalez:
@thinkyhead 
> It does not indicate that the move has been started or finished

So you think sending a "movement complete" response after any G0/1 is done (leaves the block_buffer) would have a significant performance hit?


> Actually if the host disconnects and reconnects it re-initializes the printer. So there's no need to physically unplug or hit the power switch.

This is only true if the DTR line is triggered. Regardless, resetting the MCU is not a valid way to stop a print and should only be used in emergency stop circumstance.


> But the command is still queued immediately afterward, then executed once again when the queue is processed.

Although likely undesired, the command never actually has to enter the queue to begin with if we don't want it to. Otherwise `gcode_M112()` and `gcode_M108()` could be left empty too; whatever works best.

thinkyhead:
> So you think sending a "movement complete" response after any G0/1 is done (leaves the block_buffer) would have a significant performance hit?

So, every move handed to the planner would also include the "line number" of the command (if there is one)? Then every `G0`/`G1`/`G2`/`G3` will sends two responses back to the host? It might start to add up. The stepper interrupt will keep on keeping on, and the command queue will have no trouble staying ahead of it. But I don't know if it's worth spending the extra electrons for … what extra benefit? Once a move command enters the planner queue it's not the host's problem any more.

scalez:
@thinkyhead I'm not a host guy, @kakaroto would understand the benefits much more than I would, I just wanted to understand your point of view.

kakaroto:
@repetier I didn't know repetier sends line numbers, in that case the line number sent during printing will be a sequential number, and not the actual line number of the gcode file, is that right? and I assume you keep track of the mapping between line number sent to marlin vs line number in the file (as well as custom commands with their associated line numbers) in case of a resend? That's a good idea I guess, although it adds quite a bit of complexity on the host.

As for how you keep track of bytes sent, etc.. We thought about that but I think it makes things extremely difficult and mostly, very easily prone to errors, like you said. Also, it's not needed if you only keep track of commands sent, and make sure you don't send more than the queue size (BUFSIZE in marlin, set to 4 by default). The number of bytes in the command queue is irrelevant (3 commands will take 3 slots, regardless of the number of bytes they contain), it only makes sense when you start filling the serial buffer, and that shouldn't happen as long as the command queue is not full. It also makes it much more safe against 'ok' sent in the wrong order. So it works this way without adding the complexity and while being more robust than keeping track of bytes sent. You might want to consider doing that.
the 'ping pong' mode is safer but disables look-ahead that the firmware needs to have better prints.  

> So as I understand the new system breaks order of "ok" under some circumstances and makes it impossible to count correctly since we can not match them to the line we send. That might cause a comm error with triggered resend to fix it. But the number of these commands is low and rarely used, so we could ignore that problem.

If you're referring to M108, then no, it only gets processed right away, but it still enters the queue and get its 'ok' sent later when it gets re-processed later in the correct order.


@thinkyhead 
> No, the "ok" response is sent according to the same policy as every other command. The G1 command is considered "processed" as soon as the move has been added to the queue. The "ok" response indicates that the command has been understood, accepted, and the move is now queued in the planner. It does not indicate that the move has been started or finished (try making a firmware that works that way and see how it performs). This is not a Marlin thing. It's the way a real-time robot firmware must work.

Current Marlin, yes, but old Marlin (about a year ago I think) was sending the 'ok' in get_command for G0 to G3 commands, so the 'ok' was sent asynchronously, and when the G1 command gets processed (move done), it doesn't send an 'ok' because it already sent it async. I was explaining how I assume that behavior was there to allow look-ahead functionality when the host only sends 1 command at a time. That behavior was changed/fixed and the 'ok' is now sent when the G1 command is read/processed from the command queue into the planner.

> Actually if the host disconnects and reconnects it re-initializes the printer. So there's no need to physically unplug or hit the power switch.

It does, but that's because the USB to serial firmware (on M32U2 for RAMBo) does it, but it's not guaranteed and nothing indicates which hardware is used or how it behaves on new connections or even whether or not USB to serial is used, could be a direct UART access, in which case it wouldn't work without doing a hardware reset (using DTR).

> So, every move handed to the planner would also include the "line number" of the command (if there is one)? Then every G0/G1/G2/G3 will sends two responses back to the host? It might start to add up. The stepper interrupt will keep on keeping on, and the command queue will have no trouble staying ahead of it. But I don't know if it's worth spending the extra heat and electrons for … what extra benefit? Once a move command enters the planner queue it's not the host's problem any more.

No, just the way Marlin currently works is good. Once it enters the planner, it can send the 'ok' and it's all good. I was referring to this specific bug actually : https://github.com/MarlinFirmware/Marlin/issues/1147 which was fixed with this: https://github.com/MarlinFirmware/Marlin/commit/acc0e7527914948656ccabba35f7faedc94ef885 in December 2014. I believe that's what @repetier referred to as well as being a problematic issue when we want to track commands in the host.

scalez:
@thinkyhead since the "ok" is sent when the `G0 / G1 / G2 / G3` command enters `block_buffer` and frees an element in `commands_in_queue` as fixed by @foosel https://github.com/MarlinFirmware/Marlin/issues/1147 then the `commands_in_queue` buffer is behaving as we'd expect. We were just used to seeing the old behavior for a while, sorry about that.

repetier:
@thinkyhead Thanks for the clarifications. So ordering of "ok" is kept in tact and everything is therefore backward compatible to all hosts as far as I can say..

Here my comments.

line numbers: We send line numbers for each command and have a history of last 50 commands or so so in case of resend we know what it was. This has the benefit that all commands get executed. After all the reason for line numbers is to allow resending of wrong commands and detect missed lines. It was so evident to me that at one point I even disallowed missing line numbers in firmware when you started using line numbers. But only until I learned some hosts do not always send line numbers:-)

Line numbers are also the best way to make sure buffer filling is correct. I'm not sure everyone understood so here some examples how we handle this. Lets assume we know 3 buffers are in place (or input buffer can always hold 3 commands, which is the same):
Send 
N1 M105
N2 M105
N3 M105
Return
ok 1
rk 2 ; communication error
ok 3
If I count ok now only 2 buffers get filled. If I see ok 3 and see line 2 as unacknowledged I know I missed 2 (which triggers Missed line detected message) and we pop 2 ok insted of 1. That way we can correct 1-2 missed "ok" if line numbers are used. 

If we had
M105
M105
N1 M114
ro  ; not detected 3 buffer filled assumed
ok ; pop 2 buffer filled
ok 1 ; good luck, we see a ok without line befoe our line so we pop 2 and are in sync

So as long as we get a ok with line number before we get "buffer size" undetected "ok" this can get us in sync with full buffers.

Currently we have 2 different kinds of firmwareas as it looks. The non buffering firmwares like old Marlin, Repetier, ... and the new firmware having a internal command buffer. I will not say the buffer solution is bad. In fact the input buffer count trick does more or less simulate that to get the same benefits even for firmwares not having buffers. Host problem is to distinguish both. It is completely out of question to ask users. I've done that multiple times and the often don't even know the basic version, so asking if their version support internal buffers and how many they have is not a real solution. What I do at connection is to post a number of query commands to ask for firmware capabilities. So is there a command or output that clearly says that buffers are used and how many there are? That would make it possible to automatically switch handling.

sending "ok" for G0-3 when they are in planner is always ok and also assumed. This also does not break order of ok and also does not really interest hosts as hosts are only interested in sending data. And since order is preserved this should work with all hosts as far as I can see.

scalez:
@Blue-Marlin can you please re-post that comment here?

Blue-Marlin:
An idea for a stateful parser in the RX interrupt to detect M112/M108 before they enter the RX-buffer:

----
```
    FORCE_INLINE void checkRx(void) {  // the 'hack'. Called from stepper-ISR, because RX-ISR is blocked during stepper-ISR.
      if (TEST(M_UCSRxA, M_RXCx)) {
        unsigned char c  =  M_UDRx;

// check/parse here !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!

        CRITICAL_SECTION_START;
          uint8_t h = rx_buffer.head;
          uint8_t i = (uint8_t)(h + 1) & (RX_BUFFER_SIZE - 1);

          // if we should be storing the received character into the location
          // just before the tail (meaning that the head would advance to the
          // current location of the tail), we're about to overflow the buffer
          // and so we don't write the character or advance the head.
          if (i != rx_buffer.tail) {
            rx_buffer.buffer[h] = c;
            rx_buffer.head = i;
          }
        CRITICAL_SECTION_END;
      }
    }
```
```
FORCE_INLINE void store_char(unsigned char c) {

// and check/parse here !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!

  CRITICAL_SECTION_START;
    uint8_t h = rx_buffer.head;
    uint8_t i = (uint8_t)(h + 1)  & (RX_BUFFER_SIZE - 1);

    // if we should be storing the received character into the location
    // just before the tail (meaning that the head would advance to the
    // current location of the tail), we're about to overflow the buffer
    // and so we don't write the character or advance the head.
    if (i != rx_buffer.tail) {
      rx_buffer.buffer[h] = c;
      rx_buffer.head = i;
    }
  CRITICAL_SECTION_END;
}
```
At that places the check/test works even when the RX-buffer is full and the chars are thrown away in the next if statement.

The same place for the AT90USBs is in https://github.com/MarlinFirmware/Marlin/blob/3ba9b06b798b626db4012b1010c068ab3aa4e002/ArduinoAddons/Arduino_1.6.x/hardware/marlin/avr/cores/at90usb/CDC.cpp#L141-L155

Testing single CTRL-chars would be fastest, but sending them has to be supported by the hosts. CRTL-chars because they do not appear in the normal communication.

If a little statefull parser is fast enough has to be tested.
```
enum pstates = reset, M, M1, M10, M11;

switch (state)
  case reset: if( inputchar == 'M' ) state = M;

  case M:  if( inputchar == '1' ) state = M1;
           else state = reset;

  case M1: switch (inputchar)
             case '0': state = M10;
             case '1': state = M11;
             default: state = reset;

  case M10: if( inputchar == '8' ) state = reset; setflag for M108;

  case M11: if( inputchar == '2' ) state = reset; setflag for M112;

  default: state = reset;
   
```

jbrazio:
[repost from the other thread]

This is an interesting idea, what what happens to each byte ?
Let's imagine we are really receiving a "M112", the state var will be:
```
checkRx() 1: inputchar == 'M' -> state = M
checkRx() 2: inputchar == '1' -> state = M1
checkRx() 3: inputchar == '1' -> state = M11
checkRs() 4: inputchar == '2' -> state = M112 -- trigger();
```
But int he end of the three first iterations `rx_buffer.buffer[h] = c;` will still be called right ? This means that RX-buffer will have "M11" when it gets a trigger (assuming we trigger before storing the byte of the 4th iteration), thus somewhere upfront we have to remove "M11"/"M112" from the buffer ?

The main issue here is scalability, I understand we shouldn't have so many "immediate commands" but I can see the complexity of this growing exponentially.

Blue-Marlin:
@kakaroto 
Right. A test for '\n' makes sense.

@jbrazio 
There is no state M112. If M112 is detected the state is reset again. 
Scalability is not really an issue. If the state machine grows to infinity we could use a lex/flex parser. But that is far far away.

The commands are still feed into the RX-buffer and can be handled the way we want at the place we want.

jbrazio:
> There is no state M112

No that's the var name you have used on your example. My `trigger()` is yours `setflag for M112;`.

This flag thing goes somehow into #379 direction.

Blue-Marlin:
@jbrazio 
Not really. A common state machine for the complete Marlin would be much to complex to maintain and would have hundreds of states. Separate state machines for objects/sub-objects are ok for me. Setting a flag is only an idea from the original PR (cancel_wait_heatup).

AnHardt:
Testable preview including M112, M410, M108: https://github.com/AnHardt/Marlin/tree/miniparser-in-MarlinSerial

AnHardt:
Tests with Pronterface worked well so far.
But testing with RepetierHost sowed - it's sending all lines with line numbers - so requiring a '\n' to start is not a good idea.

repetier:
What do you mean with \n to start? Shouldn't \n end each line?
Sending line numbers should be no problem - we do this for book keeping and all commands could also be part of a program where all hosts send it with line numbers.

M117 is always a pain. Some marlin versions need no line number, some can not handle ":" in it and some do not allow checksum at the end.

AnHardt:
@repetier 
Just a logic error by us.

The parser has grown a bit now, but has still the same speed. Accepts M117 and comments (';') now .

Still testing.

AnHardt:
@repetier 
I try to find a way to catch M112 and similar commands  _before_ they are put into the RX-buffer of the serial. The effect i hope for, is to be able to send M112 an co. at any time - regardless how full the buffers are.

repetier:
Wasn't the discussion about M108 about the same? So your problem is that M117 or file names with M112 may cause false emergency stops I guess. So the \n makes sense now.Guess you need to select the content between two \n (\r might be send as well my some hosts but could be converted to \n on the fly) and then do some sanity checks if M112 is contained.

AnHardt:
@repetier 
Filenames! You name it.
M108? Yes, this is still a follow up of this thread.

Courrent test with Probterface do work. With RH, works when not in pingpong-mode. (Who wonders - no 'ok' no new line). Octoprint seems to have a filter for M112 and does send it in V1.2.11, but not M108 or M410.

repetier:
Even without ping pong it depends on buffer size. But M112 is short so it gets send often or with next freed command. I'm also thinking about special handling to send it anyway for next release. Repetier-Server should work similar to host. Also it will try to reset after firing M112 to speed up things:-) Faster way wins.

scalez:
@AnHardt 
Awesome, I like the `EMERGENCY_PARSER` macro :+1:. Maybe we can echo a message to the host during the `setup()` function which let's it know that it's enabled.

There is no way for Marlin to prevent commands from entering any kind of buffer (`rx_buffer` or prior) without potentially being in a situation where *all* buffers are full and asynchronous commands *may* be sent. In that situation, we would have to throw away the synchronous commands and request re-sends (let the host know we're not holding it).

I think it's fair to say that if the host fills the command buffer and is willing to send more commands whilst the command buffer is full (filling `rx_buffer`) then it is misbehaving. If `rx_buffer` overflows then commands *will* be truncated and lost. We're handing this by echoing a message which tells the host that the `rx_buffer` has been filled if/when it happens: https://github.com/alephobjects/Marlin/commit/dad901797cc311f74d8020d2eb41180ecf759f15

I propose we expand `BUFSIZE` for certain MCUs/`MOTHERBOARD`s (e.g. ATmega2560) to give hosts enough headroom for Marlin to be able to properly handle asynchronous commands. This, coupled with `ADVANCED_OK` would let hosts be confident that Marlin will properly handle async calls.

AnHardt:
@scalez 
> Maybe we can echo a message to the host during the setup() function which let's it know that it's enabled.

We could extend ``M115`` by, for example, "EMERGENCY_CODES:M112,M108,M410;"

> There is no way for Marlin to prevent commands from entering any kind of buffer (rx_buffer or prior) without potentially being in a situation where all buffers are full and asynchronous commands may be sent. In that situation, we would have to throw away the synchronous commands and request re-sends (let the host know we're not holding it).

Please reread my code. The ermergency command parser does not need the RX-buffer. The history is stored in the states. It does not matter if we can store the 'asynchronous commands' or not, because they are executed immediately when seen. In the normal parser they are not checked anymore and simply return with an 'ok'. 
The pure knowledge, of the host, about if we have thrown away some chars because the RX-buffer was full, does not help him to decide what to resend. At the other side of the RX-buffer, when we are in the normal parser, when we check linenumbers and checksums, we can tell the host what to resend. The information about, if we have had a buffer overflow or not is interesting just for debugging purpose.

> alephobjects/Marlin@dad9017

Seems to be very untested. ':'? ';'! Sending error in the SCARA code? Up to 8 decisions in the interrupt code? Cleaning up the buffer backwards means messing up the checksum. I can't find any good idea there in.

In a good host program (like RH) you can adjust the RX-buffer size. Telling the host you have a bit less buffer than you really have makes sense. RX-buffer sizes different from powers of 2 are awfully slow to calculate in Marlin. The next step in Marlin would be 256 bytes - a bit to much for my taste. Room for one long line of code should be enough. 

`ADVANCED_OK` in the current form is very useful for debugging, but not for the host. The host do not need to know about the planner buffer. Much more interesting could be the RX-buffer, what is not in there.

scalez:
>We could extend M115

:+1: 

> Sending error in the SCARA code?

@AnHardt this was a mistake, fixed in a following commit https://github.com/alephobjects/Marlin/commit/65e4ef99aeae63348ce19dd4027bdabb7e507681. It's been tested many times; when the command buffer is filled the message is sent.

>In a good host program (like RH) you can adjust the RX-buffer size.

:+1:

boelle:

scalez:
@boelle thanks Bo, we have a PR against RCBugFix here: https://github.com/MarlinFirmware/Marlin/pull/3611, could you please reopen it?

boelle:
done---... closing this one

